PR #2, opened 2019-08-06
D8CORE-139 Basic Page Content Type

pookmish:
# READY FOR REVIEW

# Summary
- Basic page content type with a paragraphs field to get started.
- using pacman layout which doesnt work correctly right now. work will be needed for that
- rebuilt all config to start with minimal profile. This eliminated a bunch of unwanted stuff from the standard profile

# Needed By (Date)
- 8/16

# Urgency
- medium

# Steps to Test
1. follow https://github.com/SU-SWS/acsf-cardinalsites/pull/11

# See Also


config/install/node.type.su_basic_page.yml
```diff
@@ -0,0 +1,18 @@
...
+dependencies:
+  module:
+    - menu_ui
+third_party_settings:
+  menu_ui:
+    available_menus:
+      - main
+    parent: 'main:'
+name: 'Basic Page'
+type: su_basic_page
```
sherakama: Please have a review of the naming conventions document: https://docs.google.com/spreadsheets/d/1ATP3boyFio_s2BYjvzBoOt29p7aQrQV3HB-kZOZwp0g/edit#gid=0

@boznik please also have a review of the naming conventions document.

config/install/core.date_format.fallback.yml
```diff
@@ -0,0 +1,10 @@
+uuid: 0dda3437-336e-4a68-addf-e97fe1732c76
```
sherakama: We have a date format module that we should continue to standardize and store these in. 
https://github.com/SU-SWS/stanford_date_formats/tree/8.x-1.x

config/install/core.entity_form_display.media.file.default.yml
```diff
@@ -0,0 +1,36 @@
+uuid: b22c5674-c1a0-4694-b296-1c208785d0fa
```
sherakama: I am assuming that these came from the stanford_media module?

config/install/core.extension.yml
```diff
@@ -0,0 +1,96 @@
+module:
+  admin_toolbar: 0
+  admin_toolbar_links_access_filter: 0
+  admin_toolbar_tools: 0
+  allowed_formats: 0
+  anchor_link: 0
+  big_pipe: 0
```
sherakama: Pipe? Really?

config/install/core.extension.yml
```diff
@@ -0,0 +1,96 @@
...
+  jumpstart_ui: 0
+  layout_builder: 0
+  layout_discovery: 0
+  link: 0
+  linkit: 0
+  media: 0
+  media_duplicate_validation: 0
+  menu_block: 0
+  menu_ui: 0
+  nobots: 0
```
sherakama: Perhaps leave nobots out of the core install and leave it up to the environment to manage.

config/install/image.style.large.yml
```diff
@@ -0,0 +1,17 @@
+uuid: 74996e0d-0fd0-4db9-8f72-7b0725ed7ce4
```
sherakama: We have an image styles module to share a number of common styles: https://github.com/SU-SWS/stanford_image_styles

I am open to dropping the module as it tends to couple things over the long term.

config/install/node.type.su_basic_page.yml
boznik: got it, should be stanford_basic_page

config/install/search.page.user_search.yml
```diff
@@ -0,0 +1,14 @@
+uuid: c64dd58b-d2e9-45e1-8f51-662ed024a84a
```
sherakama: We don't want a user search page.

stanford_profile.info.yml
```diff
@@ -23,16 +23,13 @@ install:
   - drupal:field_ui
   - drupal:file
   - drupal:help
-  - drupal:history
-  - drupal:honeypot
```
sherakama: No honeypot?

stanford_profile.info.yml
```diff
@@ -23,16 +23,13 @@ install:
...
   - drupal:image
   - drupal:menu_link_content
   - drupal:menu_ui
   - drupal:node
   - drupal:options
   - drupal:path
   - drupal:page_cache
-  - drupal:rdf
```
sherakama: Why not rdf?

config/install/core.entity_form_display.media.file.default.yml
pookmish: yes

config/install/core.extension.yml
pookmish: why not? better performance

config/install/core.extension.yml
pookmish: i put it into there since we arent going to turn nobots off until RC1. we could move it then

stanford_profile.info.yml
```diff
@@ -44,32 +41,17 @@ install:
   - admin_toolbar_links_access_filter:admin_toolbar_links_access_filter
   - admin_toolbar_tools:admin_toolbar_tools
   - config_split:config_split
-  - profile_split_enable:profile_split_enable
   - allowed_formats:allowed_formats
   - chosen:chosen
-  - coffee:coffee
```
sherakama: I find this module helpful. Can we keep it or replace it with something similar.

config/install/image.style.large.yml
pookmish: these three came from core image module

config/install/search.page.user_search.yml
pookmish: oh that one snuck in there.

stanford_profile.info.yml
pookmish: No forms, no need for honeypot

stanford_profile.info.yml
pookmish: No need at this time

stanford_profile.info.yml
```diff
@@ -44,32 +41,17 @@ install:
...
-  - honeypot:honeypot
   - pathauto:pathauto
   - redirect:redirect
-  - webform:webform
-  - xmlsitemap:xmlsitemap
   - nobots:nobots
 
 dependencies:
   # Stanford functionality.
-  - su-sws:stanford_date_formats
```
sherakama: Are we removing these with the intention to add them back in on the appropriate ticket or what?

stanford_profile.info.yml
pookmish: toolbar module has the same functionality, just without the keyboard shortcut

stanford_profile.info.yml
pookmish: yeah. this is just very boiler plate install that we'll expand on as we go. i was getting conflicts with these lines. I kinda think its because of the `su-sws` prefix. during install it kept saying "not found". we might have to investigate the prefix later

stanford_profile.install
```diff
@@ -22,8 +22,9 @@ function stanford_profile_install() {
 
   // Assign user 1 the "administrator" role.
   $user = User::load(1);
-  $user->name = 'sws-developers';
-  $user->roles[] = 'administrator';
+  $user->set('name', 'sws-developers');
+  $user->set('mail', 'sws-developers@lists.stanford.edu');
```
sherakama: I am good with hard coding this. I don't think we will ever want to hand over user 1 to any of our clients.

config/install/node.type.su_basic_page.yml
pookmish: can we review the naming convention on fields more. field config & field storage config are namespaced `field.` already. Is there a need to have `field_s` before every field

boznik: good point

stanford_profile.install
```diff
@@ -69,6 +70,41 @@ function stanford_profile_install() {
...
+  user_role_grant_permissions(RoleInterface::ANONYMOUS_ID, [
+    'access content',
+    'search content',
+  ]);
+  user_role_grant_permissions(RoleInterface::AUTHENTICATED_ID, [
+    'access content',
+    'search content',
+  ]);
+
+  if (\Drupal::moduleHandler()->moduleExists('default_content')) {
```
sherakama: I seem some additional error handling, but why was the content import moved to the install rather than listed in the .info.yml?

stanford_profile.install
pookmish: thing to note. `blt drupal:install` overrides these. but `drush si config_installer` uses them. I think when installing through ACSF UI it will work correctly

config/install/node.type.su_basic_page.yml
sherakama: You both have been assigned a review of that document. Please review and leave comments.

stanford_profile.install
pookmish: error handling shows messages in the function that is called.. the lines in info.yml dont do anything. In fact by using `config_installer` default content doesnt get created. We have to explicitly call to install them

config/install/core.date_format.fallback.yml
pookmish: This is installed as part of `system` https://git.drupalcode.org/project/drupal/tree/8.8.x/core/modules/system/config/install

config/install/image.style.large.yml
pookmish: https://git.drupalcode.org/project/drupal/tree/8.8.x/core/modules/image/config/install

stanford_profile.info.yml
pookmish: theres a ticket to make more use of google assistant and alexa which will involve a whole lot of metadata work. we could wrap it into that one

config/install/core.extension.yml
sherakama: How does this play with varnish?

https://www.drupal.org/docs/8/core/modules/big-pipe/bigpipe-environment-requirements#varnish
https://docs.acquia.com/acquia-cloud/performance/bigpipe/

Looks like Acquia recommends the https://www.drupal.org/project/big_pipe_sessionless module to help with anon traffic. This is likely something we will want to keep "hits" down.

config/install/core.extension.yml
sherakama: Ok, I can get on board with that, but one last question. Won't we want to have this as part of our environment workflow stuff that we are figuring out anyway? eg: Enable/Disable on specific environments?

stanford_profile.info.yml
sherakama: Fair enough.

config/install/core.extension.yml
pookmish: Hmmm..
>>> So, when using Varnish: the first request will be streamed by BigPipe, and not cached in Varnish. The second request will be a Page Cache hit, not a Varnish hit. The third request and all later requests will be Varnish hits.

stanford_profile.info.yml
sherakama: But, but, it is the keyboard shortcut that I want.

config/install/node.type.su_basic_page.yml
boznik: i left 2 comments, one adjusting the machine name and another agreeing that it would be nice to not add another field_ to the field config and storage name.

stanford_profile.info.yml
pookmish: https://www.excelcampus.com/vba/keyboard-shortcut-run-macro/ :trollface:

config/install/core.extension.yml
pookmish: in the `acsf-cardinalsites` PR there is enable/disable on local vs ACSF so it would be an easy move for now... but when we get to the point of RC1 we dont plan to use config management so then this config will be removed at that time anyways

config/install/node.type.su_basic_page.yml
sherakama: We have the ability to customize the prefix so I would not expect us to create anything such as `field_field_s_thing` but rather just `field_s_thing`

sherakama: Where did you leave your comments?

boznik: above

pookmish: https://github.com/SU-SWS/stanford_profile/pull/2/files#diff-c508ab451427faf121fbf79be6af723eR1
this changes `field_` to `su_` because a field config and storage has `field.field.[entity_type]` and `field.storage.[entity_type]` do we really need more indication its a field? when the name gets too long, the revision table is a hashed string

config/install/core.extension.yml
sherakama: Lets pull big pipe out and put in a new ticket to explore this. We can't afford to have several hits per page on our factory subscription.

pookmish: I've been using it on swshumsci and havent seen a spike in page requests

pookmish: although i do need to review the logs more

pookmish: we can investigate later.

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-329

config/install/node.type.su_basic_page.yml
sherakama: I agree with those points, and the hashed strings are annoying but there are benefits to being the same as everyone else. We should discuss.

config/install/core.extension.yml
jbickar: I would vote to put this logic in Apache.

jbickar: Nobots was written for users who do not have access to Apache config and only to the Drupal GUI.

config/install/node.type.su_basic_page.yml
sherakama: discussed .`su_` it is.

config/install/core.extension.yml
pookmish: but we'll also need nobots for non-production sites on the production. It still gonna be needed

jbickar: Good call. We want the module there but disabled. General logic for `dev/test/prod` in `.htaccess`. User-controlled `nobot` for `prod`.

config/install/chosen.settings.yml
```diff
@@ -1,16 +1,16 @@
...
-minimum_multiple: 0
+minimum_single: 20
+minimum_multiple: 20
 disable_search_threshold: 0
-minimum_width: 200
-jquery_selector: 'select:not([name*="[display_id]"])'
-search_contains: true
+minimum_width: 0
+use_relative_width: false
+jquery_selector: 'select:visible'
```
sherakama: Aggressive, but good for now.

config/install/components.settings.yml
```diff
@@ -0,0 +1,3 @@
+namespace_prefix: 0
```
sherakama: From the module:

>          // Prefix the namespace with original name.
>          // Prevents overriding components with duplicate valid namespaces.

Might be worth looking at later.

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-331

config/install/core.entity_view_display.node.stanford_page.default.yml
```diff
@@ -0,0 +1,103 @@
...
+    - layout_builder
+    - layout_discovery
+    - user
+third_party_settings:
+  layout_builder:
+    allow_custom: false
+    enabled: true
+    sections:
+      -
+        layout_id: layout_onecol
```
sherakama: Do we need this section? If so, can we switch to the `basic` layout in jumpstart_ui so we match the responsive grid settings?

pookmish: oh i think this just needs to be removed. i'm using the pacman_left layout below this line. which doesnt work as a dynamic layout. the left sidebar is always visible even if no content exists. But we can visit that later.

config/install/components.settings.yml
pookmish: interesting....
``` if (!empty($config->get('namespace_prefix'))) {
              $namespace = "{$name}_{$namespace}";
            }

            $this->libraries[] = array(
              'type' => $type,
              'name' => $name,
              'namespace' => $namespace,
              'paths' => $paths,
              'error' => FALSE
            );```

File: config/install/chosen.settings.yml (hunk 14 above)
Comment (pookmish) [reply config/install/chosen.settings.yml]: yeah we'll adjust as we need

File: config/install/components.settings.yml (hunk 15 above)
Comment (pookmish) [reply config/install/components.settings.yml]: so basically all that this does is namespace the library with the module/profile/theme name. So if jumpstart_ui & stanford_basic have the same `component-libraries` defined in the info.yml, then the libraries would be prefixed with the extension name. for example is both the module and the theme have `decanter` `component-libraries` as:
```
component-libraries:
  decanter:
    paths:
      - patterns
```

Then the each extension has its own namespaced library as `jumpstart_ui_decanter` and `stanford_basic_decanter`

File: config/install/entity_usage.settings.yml [hunk 17]
```diff
@@ -0,0 +1,5 @@
+track_enabled_base_fields: false
```
sherakama: > If enabled, relationships generated through non-configurable fields (basefields) will also be tracked.

I'm not grokking this at the moment.

sherakama: /contrib/entity_usage/src/EntityUsageTrackBase.php 

`getReferencingFields` method

pookmish: I think thats if a base field such as `title` or `changed` have a entity reference, it wont track those. i see no circumstance we'd ever want/need that

sherakama: yeah, that is my foggy understanding of this as well. Default is false and I have been happy with the usage reports so far.

config/install/field.field.user.user.field_user_display_name.yml
```diff
@@ -0,0 +1,22 @@
+uuid: 82d09ddd-02da-4ef6-a6b5-1718ab447ed3
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.user.field_user_display_name
```
sherakama: This field name will need to change with the new conventions.

pookmish: that shouldnt even be there yet.

config/install/field.storage.user.field_user_display_name.yml
```diff
@@ -0,0 +1,23 @@
+uuid: a4fc9093-dd68-4412-bb71-60eedad294e1
+langcode: en
+status: true
+dependencies:
+  module:
+    - user
+_core:
+  default_config_hash: 9p-e2WTkPJrF3t_8DRzPFr8VzxLfhFXo-1865eGbXtA
+id: user.field_user_display_name
```
sherakama: new `su_` prefix for user fields.

config/install/responsive_image.styles.full_responsive.yml
```diff
@@ -0,0 +1,37 @@
...
+    - image.style.responsive_small
+  module:
+    - stanford_media
+_core:
+  default_config_hash: 87YDNYUvaTv6bZRJVF4DsV7apqKNcQIY5xTHMC97ZjM
+id: full_responsive
+label: 'Full Responsive'
+image_style_mappings:
+  -
+    breakpoint_id: stanford_media.xl
```
sherakama: These will likely need to be re-mapped to match the breakpoints in jumpstart_ui/stanford_basic

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-334

config/install/system.file.yml
```diff
@@ -0,0 +1,7 @@
+allow_insecure_uploads: false
+default_scheme: public
+path:
+  temporary: ''
```
sherakama: We will have to remember to set this for the LB and webheads so that it doesn't just default.

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-335

config/install/system.image.yml
```diff
@@ -0,0 +1,3 @@
+toolkit: gd
```
sherakama: Acquia recommends imagemagik: 
https://support.acquia.com/hc/en-us/articles/360005255734-Scoping-the-impact-of-image-processing#usetherightimagepackage

Explore this change.

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-336

config/install/system.site.yml
```diff
@@ -0,0 +1,15 @@
+uuid: 9ef76ae9-9319-47b7-9cf4-4ebd82a0d3d6
+name: 'Drush Site-Install'
+mail: admin@example.com
+slogan: ''
+page:
+  403: ''
```
sherakama: We should create these.

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-337

config/install/taxonomy.settings.yml
```diff
@@ -0,0 +1,5 @@
+maintain_index_table: true
```
sherakama: TIL you can turn off the index table. 

> * Taxonomy uses default field storage to store canonical relationships
> * between terms and fieldable entities. However its most common use case
> * requires listing all content associated with a term or group of terms
> * sorted by creation date. To avoid slow queries due to joining across
> * multiple node and field tables with various conditions and order by criteria,
> * we maintain a denormalized table with all relationships between terms,
> * published nodes and common sort criteria such as status, sticky and created.

sherakama:
@boznik & @jbickar, you're up.

config/install/taxonomy.settings.yml
pookmish: well thats interesting. nice find

boznik:
did you want to change the name here to just page?
https://github.com/SU-SWS/stanford_profile/blob/D8CORE-139-basic-page/config/install/node.type.stanford_page.yml#L12

pookmish:
Meh, we can change the label easily if UX'ers think necessary

jbickar:
I did not get the chance to review this. Don't let me be a blocker. I have faith in @sherakama and @boznik.

This comment has a persistent URL if Future Us needs to blame Today John.

pookmish:
I'm pulling the trigger. its nothing we cant change in the future

PR #3, opened 2019-08-12
D8CORE-143 Text formats and WYSIWYG paragraph

pookmish:
# READY FOR REVIEW

# Summary
- WYSIWYG text editors and formatters:
  * HTML: normal html for pretty much everything
  * Minimal HTML: Very stripped down HTML for use in captions and simple fields.

# Needed By (Date)
- End of the week.

# Urgency
- Medium

# Steps to Test
- Follow steps on https://github.com/SU-SWS/acsf-cardinalsites/pull/12

# See Also


sherakama:
@pookmish, I'd like the input formats and the wysiwyg profiles to be split out into their own module. We have https://github.com/SU-SWS/stanford_text_editor that could be a home for them, otherwise, please stub out a module in this profile.

pookmish:
I thought the plan was to build it out and break it into modules later? Creating a module inside the profile would be nothing but duplicate configs

sherakama:
For the majority of our work yes, for a few items that have been identified as something that will be shared for everything we may as well do the abstraction from the start. Those modules were:

* Media
* Text formats and wysiwyg profiles
* SSP
* Jumpstart UI - layouts/components

pookmish:
so because we're using config management, moving the format and wysiwyg profile to the other module might only change 1 thing in this pr: the `core.extensions.yml`. so this PR can still be reviewed in the mean time I'd say

pookmish:
The down fall of breaking out the configs into another module. now we have that order of operations issue https://github.com/SU-SWS/stanford_profile/pull/3/commits/fdd987b76ec4478a0caf56ffa83810a97e235c6c

tests/behat/Paragraphs/WYSIWYG.feature
```diff
@@ -0,0 +1,153 @@
...
+    And I am on "/node/add/stanford_page"
+    Then I fill in "Title" with "Test WYSIWYG Paragraph"
+    And I press "Add WYSIWYG"
+    And I wait for AJAX to finish
+    Then I click the "a[title='Embed Media']" element
+    And I wait for AJAX to finish
+    Then I switch to "entity_browser_iframe_media_browser" iframe
+    Then I click "Embed a File"
+    And I wait for AJAX to finish
+    Then I drop "../assets/images/logo.jpg" file into dropzone
```
sherakama: I am getting a Javascript error "Dropzone is undefined" on this step.

sherakama: Seems that the step `And I wait for AJAX to finish` isn't waiting long enough.

tests/behat/Paragraphs/WYSIWYG.feature
```diff
@@ -0,0 +1,159 @@
...
+    Then I click the "a[title='Embed Media']" element
+    And I wait for AJAX to finish
+    Then I switch to "entity_browser_iframe_media_browser" iframe
+    Then I click "Embed a File"
+    And I wait for AJAX to finish
+    And I wait 1 seconds
+    Then I drop "../assets/images/logo.jpg" file into dropzone
+    And I press "Add to Library"
+    And I wait for AJAX to finish
+    And I wait 1 seconds
```
sherakama: Not sure why the AJAX wait is not working.

tests/behat/Paragraphs/WYSIWYG.feature
```diff
@@ -0,0 +1,159 @@
...
+    Then I fill in "Title" with "Test WYSIWYG Paragraph"
+    And I press "Add WYSIWYG"
+    And I wait for AJAX to finish
+    Then I click the "a[title='Embed Media']" element
+    And I wait for AJAX to finish
+    Then I switch to "entity_browser_iframe_media_browser" iframe
+    Then I click "Embed External Content"
+    And I wait for AJAX to finish
+    And I wait 1 seconds
+    Then I fill in "Shareable Url" with "https://www.youtube.com/watch?v=ktCgVopf7D0"
```
sherakama: This item stays in the media library after the tests.

pookmish: It gets removed on my test runs.

tests/behat/Paragraphs/WYSIWYG.feature
pookmish: your machine is too slow? :trollface:

tests/behat/Paragraphs/WYSIWYG.feature
pookmish: I got that once. if we get it often, we could just put in some extra code into the step definition to wait for `Dropzone` to be available.

pookmish:
Anything we're waiting on for this?

tests/behat/Paragraphs/WYSIWYG.feature
sherakama: Slow or too fast? Not sure, but the test is trying to run the next step before the javascript has all initialized. That extra wait is for the scripts to run after the ajax call finishes. Sometimes adding the second 'and I wait for ajax' catches the new scripts loading but sometimes not.

tests/behat/Paragraphs/WYSIWYG.feature
sherakama: I'm not worried about artifacts. Tests are not to be run on production sites.

sherakama:
Should we wait for @boznik to review?

pookmish:
!(poke)[https://i.gifer.com/Q9gt.gif]

boznik:
ok, i run this through

boznik:
woohoo. behat is working

PR #4, opened 2019-08-14
D8CORE-153: Enable config_readonly

sherakama:
# READY FOR REVIEW

# Summary
- Enables config_readonly

# Needed By (Date)
- End of sprint

# Urgency
- Low

# Steps to Test

1. See https://github.com/SU-SWS/acsf-cardinalsites/pull/14

- cardinalsites stack

- D8CORE-153
- https://github.com/SU-SWS/acsf-cardinalsites/pull/14

# See Also

### DIFF
--- a/tests/behat/features/contrib/config_readonly.feature
+++ b/tests/behat/features/contrib/config_readonly.feature
@@ -6,7 +6,4 @@ Feature: Contrib - config_readonly
   @api
   Scenario: Ensure no configuration changes are possible through the UI.
     Given I am logged in as a user with the "Administrator" role
-    And the "config_readonly" module is enabled
-    # And I am on "admin/config/media/file-system"
-    # Then I check "Save configuration" button is disabled
-    # Then I should see the text "This form will not be saved because the configuration active store is read-only."
+    Then check that the "config_readonly" module is enabled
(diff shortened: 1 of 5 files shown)


tests/behat/features/contrib/config_readonly.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Contrib - config_readonly
+  In order to verify that contrib module config_readonly is working
+  As an administrative user
+  I should be not be able to administer settings when the module is active
+
+  @api
+  Scenario: Ensure no configuration changes are possible through the UI.
+    Given I am logged in as a user with the "Administrator" role
+    Then check that the "config_readonly" module is enabled
```
pookmish: wouldn't you wanna check that a form is locked. just because the module is enabled, doesn't actually lock any configs.

sherakama: I started down that path but due to the part that we don't lock config in CI environments that made it a blocker. 

The flag for enabling the locking is in the environment settings `$settings` variable which is read only and set at bootstrap. This makes it nearly impossible to set to TRUE after bootstrap. 

I am open to suggestions.

pookmish:
Missing core.extensions.yml change to enable the module

tests/behat/features/contrib/config_readonly.feature
pookmish: hmm. good point. think we'll come back to that later when we get profile CI/CD set up. then we can set the setting in something like https://github.com/SU-SWS/acsf-cardinalsites/blob/8.x-1.x/docroot/sites/settings/ci.settings.php

sherakama:
Oh, duh, thanks.

sherakama:
This gtg now?

PR #5, opened 2019-08-22
Set cache to 1 year and added saml display name field

pookmish:
# READY FOR REVIEW

# Summary
- Set cache to 1 year
- with update to stanford_ssp, new user name fields.

# Needed By (Date)
- asap

# Urgency
- medium

# Steps to Test
1. follow steps in https://github.com/SU-SWS/acsf-cardinalsites/pull/17

# See Also

### DIFF
--- a/config/install/system.performance.yml
+++ b/config/install/system.performance.yml
@@ -1,6 +1,6 @@
 cache:
   page:
-    max_age: 0
+    max_age: 31536000
 css:
   preprocess: true
   gzip: true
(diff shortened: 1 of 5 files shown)


config/install/system.performance.yml
```diff
@@ -1,6 +1,6 @@
 cache:
   page:
-    max_age: 0
+    max_age: 31536000
```
sherakama: This is overridden with the developer settings on local correct?

sherakama: Survey says... yes.

PR #24, opened 2019-10-11
D8CORE-626 Update default content

pookmish:
# READY FOR REVIEW

# Summary
- Updated default content
- Set up circleci

# Need Review By (Date)
- 10/23

# Urgency
- low

# Steps to Test
1. Checkout this branch
1. run full site install
1. validate the changes on https://stanfordits.atlassian.net/browse/D8CORE-626 are done

# See Also


composer.json
```diff
@@ -19,12 +49,16 @@
         "drupal/block_blacklist": "~1.0",
         "drupal/chosen": "~2.4",
         "drupal/components": "~1.0",
+        "drupal/config_distro": "^1.0-alpha",
```
sherakama: Do we really need these in the profile?

composer.json
```diff
@@ -54,6 +88,7 @@
         "drupal/xmlsitemap": "~1.0@alpha",
         "ext-imagick": "*",
         "su-sws/jumpstart_ui": "dev-8.x-1.x",
+        "su-sws/nobots": "dev-8.x-1.x",
```
sherakama: Does this profile rely on this module? Perhaps the stack does but not this install profile.

composer.json
pookmish: because of the core.extension.yml, yeah. the tests dont work otherwise

composer.json
pookmish: anything in core.extension.yml has to be added as a dependency for testing

sherakama: When did that get in there? Ok thanks.

composer.json
sherakama: These aren't needed by the profile but rather by the stack. We could probably pull them out of core.extension couldn't we?

composer.json
pookmish: probably BLT https://github.com/SU-SWS/acsf-cardinalsites/blob/8.x-1.x/blt/blt.yml#L31

composer.json
pookmish: we'd have to move them to the environment splits but then i'm guessing it would be in every config split. i dont think we can get around it. unless we move them to `dev-dependencies` here.

sherakama: It doesn't sound worth the move. They don't really hurt being in here so let's not fight this fight until later. We will have the need for individual sites to have different sets of modules enabled but that will be tackled in our update practices.

PR #33, opened 2019-10-25
added missing link style

pookmish:
# READY FOR REVIEW

# Summary
- Missing style makes tests fail

# Need Review By (Date)
- asap

# Urgency
- high

# Steps to Test
1. review styles

# See Also

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -97,13 +97,18 @@
         "su-sws/stanford_media": "dev-8.x-2.x",
         "su-sws/stanford_paragraph_types": "dev-8.x-2.x",
         "su-sws/stanford_ssp": "dev-8.x-1.x",
-        "su-sws/stanford_text_editor": "dev-8.x-1.x"
+        "su-sws/stanford_text_editor": "dev-8.x-1.x#a3dd26e562ff9b00e147c8abf4a599642d9234e4"
     },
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -105,5 +105,10 @@
                 "https://www.drupal.org/project/menu_link_weight/issues/2995896": "https://www.drupal.org/files/issues/2018-08-29/2995896-menu-admin-per-menu-support-fixes.patch"
             }
         }
+    },
+    "config": {
```
sherakama: This might cause problems with deploy no? Have the caches cleared and we can remove this now?

pookmish: I believe its only affects CI tests. but even with this it was still getting inconsistent results when running `composer update`

sherakama: I think I'd rather the prefer controls be passed in manually than by wiring it up here. Now that some time has passed let me try without this.

pookmish: yeah i dont know why it was doing that. hours after the last merge, it still wasn't consistent. i almost wonder if a few more commits on `stanford_text_editor` would push it out far enough. :man_shrugging:

pookmish:
double approved :heavy_check_mark: :white_check_mark:

PR #38, opened 2019-11-04
SOE prep

pookmish:
# READY FOR REVIEW

# Summary
- Ignore config/install (this profile should only be using config/sync)
- Namespace block configs with stanford_basic
- added all the dependencies to info.yml file

# Need Review By (Date)
- Soon

# Urgency
- medium

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-soe-prep`
1. run a site install
1. verify it still looks the same as before.

# See Also

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+config/install
+
 # deal with ignoring or not  bin, vendor, phpstorm files...
 .idea
 .idea/**
--- a/config/sync/block.block.stanford_basic_branding.yml
+++ b/config/sync/block.block.stanford_basic_branding.yml
@@ -6,7 +6,7 @@ dependencies:
     - system
   theme:
     - stanford_basic
-id: sitebranding
+id: stanford_basic_branding
 theme: stanford_basic
 region: header
 weight: 0
--- a/stanford_profile.info.yml
+++ b/stanford_profile.info.yml
@@ -78,10 +78,12 @@ dependencies:
   - environment_indicator:environment_indicator
   - externalauth:externalauth
   - fakeobjects:fakeobjects
+  - field_validation:field_validation
   - file_mdm:file_mdm
   - file_mdm:file_mdm_exif
   - file_mdm:file_mdm_font
   - focal_point:focal_point
+  - google_analytics:google_analytics
   - image_effects:image_effects
   - imagemagick:imagemagick
   - inline_entity_form:inline_entity_form
--- a/stanford_profile.info.yml
+++ b/stanford_profile.info.yml
@@ -5,65 +5,122 @@ type: profile
 project: Jumpstart
 core: 8.x
 
-base profile: minimal
-
-# install can be turned off later
-install:
-  # core
-  - drupal:action
+dependencies:
+  - admin_toolbar:admin_toolbar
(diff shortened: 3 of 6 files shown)


stanford_profile.info.yml
```diff
@@ -5,65 +5,124 @@ type: profile
 project: Jumpstart
 core: 8.x
 
-base profile: minimal
-
-# install can be turned off later
-install:
-  # core
-  - drupal:action
+dependencies:
```
sherakama: Are you sure you want all of these to be in dependencies? Anything listed in this section cannot be uninstalled whereas the `install` key allows you to uninstall the module at a later time (through the UI or drush). At some point, I imagine we will be offering the ability to turn on and off some modules.

pookmish: hmm. i'll have to check if changing to `install` will produce the same thing

config/sync/block.block.stanford_basic_branding.yml
```diff
@@ -6,7 +6,7 @@ dependencies:
     - system
   theme:
     - stanford_basic
-id: sitebranding
```
sherakama: I haven't tested this yet but, are we going to need an upgrade path for this or are the old blocks removed on the next config import?

pookmish: no need for an upgrade path. since the UUID's didn't change, it just gets renamed in the DB

stanford_profile.info.yml
sherakama: Did you ever find out if install was ok? I think it is going to be vital for us to use the install key as we want to allow for some managing of the modules.

https://www.drupal.org/node/2952947

pookmish: I just checked and i was able to uninstall a couple modules defined in the `dependencies` I dont think the system enforces the dependencies on the profile

sherakama: https://www.drupal.org/node/2952947

> After 8.6.0
> As of #2952888: Allow install profiles to require modules, profiles can define an explicit 'install' array in their info file in addition to the 'dependencies' array. The 'dependencies' array will be treated as a set of hard dependencies, meaning that they cannot be uninstalled because the profile depends on them. 

But it _should_ be enforcing that behavior. My ears are up on this. What are we doing that is allowing for the behaviour?

sherakama: I  should have kept reading

> To maintain backwards compatibility, profiles which only define a 'dependencies' array, but not an 'install' array, will have their list of dependencies treated like an 'install' list. This backwards compatibility layer will be removed in Drupal 9, so profile authors should separate their 'install' list from their actual dependencies as soon as possible.

sherakama: Looks like we want to go with the split today.

pookmish: interesting! good find. so we'll definitely want to split it up.

pookmish: I tried with it as `install` and it worked the same. so i switched it. should be good for future

.gitignore
```diff
@@ -1,3 +1,5 @@
+config/install
```
sherakama: I've committed this but we will talk about it a touch more.

PR #40, opened 2019-11-07
D8CORE-726 Local footer config pages and block

pookmish:
# READY FOR REVIEW

# Summary
- Local footer config pages and block placement
- Preview is available at https://github.com/SU-SWS/acsf-cardinalsites/pull/39

# Need Review By (Date)
- Sometime soon

# Urgency
- medium

# Steps to Test
1. either `composer require su-sws/stanford_profile:dev-D8CORE-726-local-footer` or checkout https://github.com/SU-SWS/acsf-cardinalsites/pull/39 and run `composer install --prefer-source`
1. run a full site install or config import
1. Edit the Local footer settings at `/admin/appearance/local-footer`
1. after saving, validate your links and content are visible in the lower region of the site

# See Also


sherakama:
https://github.com/SU-SWS/stanford_basic/pull/102

stanford_profile.profile
```diff
@@ -45,3 +45,20 @@ function stanford_profile_entity_field_access($operation, FieldDefinitionInterfa
   }
   return AccessResult::neutral();
 }
+
+/**
+ * Implements hook_preprocess_html().
+ */
+function stanford_profile_preprocess_html(&$vars) {
```
pookmish: why programmatically adding it? it gets placed via block config

sherakama: In order to stick it to the bottom of the window I need to get it out of the content wrappers.

sherakama: In order to stick it to the bottom of the window I need to get it out of the content wrappers.

pookmish: Hmm. That makes me concerned. Why mot modify the css to make it stick? Or adjust the template/add another region.

pookmish: the other thing this does is requires the local footer to display all the time. I put config pages overrides to disable the block config so it wont display. i'm very against this method

sherakama: I'm piggybacking on the approach to the global footer. With the way that it works (margin-top: auto) I am not able to get the local footer to the bottom of the page when it is wrapped in all of the wrappers that the content regions have. 

If we were to move the local footer into a block region we would have to move the global footer as well and adjust the off-dialog-canvas wrapper. 

https://api.drupal.org/api/drupal/core%21modules%21system%21templates%21off-canvas-page-wrapper.html.twig/8.5.x

Not sure I want to go down that road for footers.

pookmish: why not bring the global footer back into the canvas wrapper and modify a wrapper inside that?

worst case, throw some javascript to push it down.

pookmish: My suggestion: 
1. Move the global footer out of html template and into the page template. 
1. wrap the local footer and global footer in the same container.
1. use the concepts [here](https://css-tricks.com/couple-takes-sticky-footer/#article-header-id-3) to sticky it. 

It took me about 5 minutes and i was able to get that to work. see: https://github.com/SU-SWS/stanford_basic/pull/103

sherakama: Yeah, it looks like everything needs to be wrapped in the off-canvas div. I really dislike this as it forces us to wrap everything in a div right off the start. Perhaps we can take it a step further and get rid of the dependencies on the pattern functions with the move as well as that will allow us to install this theme without the dependency on ui_patterns. I'll move the css off of the body element and move the region to the page template. The more you wrap elements the messier the html becomes and the tougher semantics get later.

sherakama:
@pookmish ready for review again.

sherakama:
@pookmish ready to review again.

pookmish:
looks like its mostly back to what i had originally. i'll run manual check this afternoon

sherakama:
Thank yew.

PR #76, opened 2020-01-31
D8CORE-941: WYSIWYG CKEditor styles

sherakama:
# READY FOR REVIEW
- Consolidate WYSIWYG paragraph styles in to one stylesheet
- Creates a CKEDitor stylesheet for wysiwyg styles.

# Summary
- Wysiwyg style stuff

# Needed By (Date)
- Sooner the better for content authors

# Urgency
- Before end of next sprint

# Steps to Test

1. Check out this branch
2. Clear all your caches
3. Edit the home page
4. Validate the styles are available in the CSS styles drop down
5. Validate the CSS stylesheets are loaded by creating a link in a heading within the WYSIWYG Editor and confirming it is red.

- D8CORE

- D8CORE-941

# See Also


modules/stanford_profile_styles/lib/scss/admin/ckeditor.scss
```diff
@@ -0,0 +1,10 @@
+@charset 'UTF-8';
+// This stylesheet creates a css file that is attached to the CKEDITOR WYSIWYG.
+@import "../paragraph-types/wysiwyg";
+
+// This extend will duplicate the su-wysiwyg-text styles in to this stylesheet
+// and add a secondary set of children selectors under the body tag. This is an
+// acceptable amount of bloat for maintaining styles automatically.
+body {
+  @extend .su-wysiwyg-text;
```
pookmish: would this be best suited for `stanford_text_editor`?

modules/stanford_profile_styles/stanford_profile_styles.module
```diff
@@ -53,3 +53,21 @@ function stanford_profile_styles_preprocess_paragraph(&$vars) {
...
+  if (!$editor->hasAssociatedFilterFormat()) {
+    return;
+  }
+
+  $known_formats = [
+    'stanford_html',
+    'stanford_minimal_html',
+  ];
+
+  if (in_array($editor->getFilterFormat()->id(), $known_formats)) {
```
pookmish: Is this necessary? the profile knows about what filters are available. :man_shrugging:

modules/stanford_profile_styles/lib/scss/admin/ckeditor.scss
sherakama: I'd argue no because different themes and different modules can influence the look and feel and we should not assume D8Core.

modules/stanford_profile_styles/stanford_profile_styles.module
sherakama: The config is all there so, yes...

Is it necessary? No, but I don't want these styles showing up on wysiwyg editors that we aren't using to display the changed output.

modules/stanford_profile_styles/lib/scss/admin/ckeditor.scss
pookmish: maybe in the theme? we have ckeditor styles already there https://github.com/SU-SWS/stanford_basic/blob/8.x-4.x/stanford_basic.info.yml#L34

modules/stanford_profile_styles/stanford_profile_styles.module
pookmish: but the output styles could be different than what is visible in the ckeditor.

modules/stanford_profile_styles/lib/scss/admin/ckeditor.scss
sherakama: These are specific overrides just for when we use a WYSIWYG editor. We don't want to mess with the default out of the box Decanter styles that are already included in that theme's file.

modules/stanford_profile_styles/stanford_profile_styles.module
sherakama: How so?

pookmish: if the css is added to the pages but not to the wysiwyg, it produces different "what you see" results

sherakama: Yes, but where is that happening?

sherakama: Stanford Basic is still adding to the ckeditor stylesheet: https://github.com/SU-SWS/stanford_basic/blob/8.x-4.x/stanford_basic.info.yml#L34

These specific heading and font style overrides are only for when we are outputting them through wysiwyg content. The original styles should be the same outside of that scope.

pookmish: its not happening right now but it could in the future or if extending this profile. I'm not really on board with the logic but its not a blocker to me for this work.

sherakama: I don't see how this implementation would change anything about future styles not getting into the CKEditor. If you add new stylesheets within other modules or profiles you will have to add the stylesheet to CKEditor manually as well. Nothing gets in automagically. I am very concerned about long term maintenance. Could you outline a scenario where the new code would be expected to be in the WYSIWYG but isn't?

PR #78, opened 2020-02-04
D8CORE-1243 saml role mapping config page

pookmish:
# READY FOR REVIEW

# Summary
- Provided config pages for populating/overriding the simplesamlphp role mapping

# Need Review By (Date)
- 2/12

# Urgency
- medium

# Steps to Test
1. checkout this branch
1. run `drush cget simplesamlphp_auth.settings role.population --include-overridden`
1. validate its just the normal workgroup mapping
1. navigate to `/admin/config/people/simplesaml` and populate with some random string. whatever you want to test.
1. run `drush cget simplesamlphp_auth.settings role.population --include-overridden`
1. validate your string is there.
1. if you'd like, set the role mapping in the form to something like `contributor:eduPersonEnttitlement,=,itservices:webservices`
1. logout & log back in. validate you received the "Contributor" role.

# See Also

### DIFF
--- a/config/sync/config_pages.type.simplesaml.yml
+++ b/config/sync/config_pages.type.simplesaml.yml
@@ -1,0 +1,26 @@
+uuid: 565c4fc5-10a3-45c7-949f-15674ba86543
+langcode: en
+status: true
+dependencies:
+  module:
+    - config_pages_overrides
+third_party_settings:
+  config_pages_overrides:
+    adb68721-a642-41a8-a15e-0f54da3f2dac:
+      field: su_simplesaml_roles
+      delta: 0
+      column: value
+      config_name: simplesamlphp_auth.settings
+      config_item: role.population
+id: simplesaml
+label: SimpleSAML
+context:
+  show_warning: true
+  group:
+    language: false
+  fallback:
+    language: ''
+menu:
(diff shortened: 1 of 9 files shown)


sherakama:
🤔 So this is an override option and would override the itservices:webservices option if it was set in the original config? I don't think we want that here. I suspect we will want to merge this config with the defaults we provide. Wouldn't want site managers changing our roles now would we? Also, at the very least we should provide some form validation for this form as the format is archaic and would be difficult for site owners to grok and/or get correct. How much extra effort is it to get the fancy version on the config page?

pookmish:
This form isn't accessible to anyone other than admins. we need some type of configurable workgroup mapping and so we can't wrap it into the original config.
this form is just for RC1 so we can do what we need. We'll be the ones editing the config on that page. we can provide a better UI form in a later version.

sherakama:
Ok, iterate. Got it.

Do me one favor, and default the values to what is in the simplesamlphp_auth config already so we don't have any accidental loss of data. I can see us forgetting to add us when we go to add another.

pookmish:
ooh good thinkin :+1:

pookmish:
default added. g2g?

PR #80, opened 2020-02-04
D8CORE-1201 Alert when leaving an unsaved node form

pookmish:
# READY FOR REVIEW

# Summary
- Alert to the user if they are about to leave the page after editing part of the page.

# Need Review By (Date)
- 2/12

# Urgency
- medium

# Steps to Test
1. Checkout this branch
1. clear caches
1. edit/create a new node without saving
1. before saving, attempt to change the page (click on another link or try to close the tab)
1. verify an alert appears to confirm.

# See Also


modules/stanford_profile_styles/lib/js/node_form.js
```diff
@@ -0,0 +1,50 @@
...
+          for (var i in CKEDITOR.instances) {
+            if (CKEDITOR.instances[i].checkDirty()) {
+              edit = true;
+              break;
+            }
+          }
+        }
+
+        if (edit && !save) {
+          return (Drupal.t("You will lose all unsaved work."));
```
sherakama:
```suggestion
          return (Drupal.t("Are you sure you want to leave this page? You will lose all unsaved work."));
```

sherakama:
Also, what about? https://www.drupal.org/project/autosave_form or https://www.drupal.org/project/auto_save_form

modules/stanford_profile_styles/lib/js/node_form.js
pookmish: This doesn't actually display anyways. The first response on https://stackoverflow.com/questions/38879742/is-it-possible-to-display-a-custom-message-in-the-beforeunload-popup
>>> Chrome removed support for custom message in ver 51
Opera removed support for custom message in ver 38
Firefox removed support for custom message in ver 44.0 (still looking for source for this information)
Safari removed support for custom message in ver 9.1

pookmish:
I've tried those autosave type of modules in the past and i believe they didn't work well with paragraphs or something at that time. 
Some quick checks and `autosave_form` [doesnt work on new nodes](https://www.drupal.org/project/autosave_form/issues/2924006)
and `auto_save_form` [is what i thought was broken](https://www.drupal.org/project/auto_save_form/issues/3100345)

modules/stanford_profile_styles/lib/js/node_form.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+
+      addListeners();
+      // Periodically add listeners to buttons to check for activity
+      setTimeout(addListeners, 2000);
+
+      // Handle backbutton, exit etc.
+      window.onbeforeunload = function () {
+        // Add CKEditor support.
+        if (typeof (CKEDITOR) != 'undefined' && typeof (CKEDITOR.instances) != 'undefined') {
```
sherakama: Do you have to check the ckeditor itself? I believe it changes the underlying form textarea values so you could just look at all form items. Re: https://git.drupalcode.org/project/confirm_leave/blob/8.x-1.x/js/confirm-leave.js

pookmish: If i remember correctly, when the ckeditor updates the text area, it doesn't trigger listeners. 

I remember there was a reason I put this in and i think it was something like that.

sherakama: This one has a similar approach to yours with the blur on inputs https://git.drupalcode.org/project/js_confirm_pop_up/blob/8.x-1.x/js/js_confirm_pop_up.js

pookmish: one thing to remember is we can't rely on a contrib for our case. because of the react widget, we have to periodically add the listeners when the react form is visible. but we could take the code and modify it

sherakama:
Ok, please confirm the need to check the ckeditors and/or if the contrib options don't work with the react implementation. Ideally it would be nice to just include another contrib module but if they don't fit or are broken we put it here. Not sure if developers will look in a `styles` module for this but the javascript file is namespaced well.

sherakama:
Also, this isn't a big piece of functionality but if the javascript breaks then the page editing can break. Can you add a test for it? Behat or a FunctionalJavascript unit test I suspect will do the job.

pookmish:
So it turns out with very minimal adjustment to the react_paragraphs, we can use the confirm_leave module.. Closing for another PR

PR #88, opened 2020-02-06
D8CORE-369: Enable metatags and XMLSitemap on prod

sherakama:
# READY FOR REVIEW/ 

# Summary
- Enables xmlsitemap, metatag, and config_ignore
- xmlsitemap creates config items for each sitemap it creates. I went with the config ignore approach please comment on that. 

# Review By (Date)
- Feb 12th

# Urgency
- Med/High as it is needed for RC1

# Steps to Test

1. Check out this branch and D8CORE-369 branch of acsf-cardinalsites
2. Do an install
3. Validate the configuration

- D8CORE

- D8CORE

# See Also


pookmish:
Behat failing because GA is getting disabled in CI.

sherakama:
I'll look into GA. Any comments about the work on the metatag/xmlsitemap stuff or the config_ignore option?

pookmish:
Why config ignore of the xml sitemap? I don't see a current scenario we need to customize them per site

config/sync/config_ignore.settings.yml
```diff
@@ -0,0 +1,6 @@
+ignored_config_entities:
+  0: 'xmlsitemap.xmlsitemap.*'
+  2: 'webform.webform.*'
```
pookmish: We aren't going to be using webform right?

sherakama: I hope we don't open that up again but it is in our stack and a common request. It was the example default text so I included it.

sherakama:
> Why config ignore of the xml sitemap? I don't see a current scenario we need to customize them per site

XML Sitemap is like webform and it creates a config item for every sitemap. This is needed to allow each site to have its own url in the sitemap. le sigh.

pookmish:
> XML Sitemap is like webform and it creates a config item for every sitemap
Not really. It uses state to target the domain of the current site. H&S has been doing just fine without ignoring the config. During launch it require a state set:
https://github.com/SU-HSDO/suhumsci/blob/develop/blt/src/Blt/Plugin/Commands/HumsciServerCommands.php#L398

This is a requirement to function correctly. you'll see that the xml sitemap module takes the form values and sets the state https://git.drupalcode.org/project/xmlsitemap/blob/8.x-1.x/src/Form/XmlSitemapSettingsForm.php#L293

I don't believe we need config ignore at this time.

composer.json
```diff
@@ -50,7 +50,6 @@
         "drupal/chosen": "~2.4",
         "drupal/ckeditor_fixed_toolbar": "dev-1.x#1e4a8191ab289b4c91d9a0f2ac6b40e82e294908",
         "drupal/components": "~1.0",
-        "drupal/config_distro": "^1.0-alpha",
```
sherakama: We are not using this and should remove it.

sherakama:
@pookmish removed config ignore and config distro while I was at it. Tracked the default config item for xml sitemap.

composer.json
pookmish: We'd need to keep it so that it can be uninstalled correctly since it's being removed from `core.extension.yml`

config/sync/metatag.metatag_defaults.node.yml
```diff
@@ -0,0 +1,12 @@
+uuid: 470ad0e9-47ab-4346-b8ca-63167db1beca
+langcode: en
+status: true
+dependencies: {  }
+_core:
+  default_config_hash: rpwvgyEURXLz_JjgMCrkS1rUv-0k3L79BpO-ReN7fDI
+id: node
+label: Content
+tags:
```
pookmish: Could we include the tags for the header image for basic pages? it might need its own config for that node type.

composer.json
sherakama: ah good point. It is installed.

config/sync/metatag.metatag_defaults.node.yml
sherakama: What tag are you looking for?

pookmish: I was thinking at least `og:image`. if i'm not mistaken it should be something like `[node:su_page_banner:entity:su_banner_image:field_media_image:entity:url]`

sherakama: Oh, you want the open graph stuff in there. I left that out as I wasn't sure what our plans on social media integration will be. If we do image we may as well do some of the other ones. I'll see about enabling those.

pookmish: yeah google reads those OG tags too. doesn't it?

sherakama: I'm not sure but I don't see how it could hurt.

sherakama:
Ok, this should be revised and refactored into a nice bow now.

sherakama:
Behat is causing an error with the react icons?
https://1141-176381770-gh.circle-artifacts.com/0/var/www/html/artifacts/wysiwyg_paragraph-103.html

pookmish:
This looks good. i'll review the failing tests. 

One thing we might want to review is the home page default content. As we are right now, the home page doesn't use the "Banner" field. so it doesn't get the `og:image` metatag.

pookmish:
@sherakama Looks like its just the GA behat thing. This is an interesting one, because the behat should test for the GA here, but not at the stack level. :thinking:

sherakama:
> @sherakama Looks like its just the GA behat thing. This is an interesting one, because the behat should test for the GA here, but not at the stack level. 🤔

Ugh, I don't want to get into testing with different tags again but perhaps we need them. By default behat tests run with the tags `~ajax&&~experimental&&~lightningextension'` but we can modify them.  

TLDR; Add the @experimental tag to things that can be overridden by settings.php config.

sherakama:
Green lights first try!

pookmish:
boom baby

PR #112, opened 2020-02-21
D8CORE-1357: Adds validation to config form uri.

sherakama:
# READY FOR REVIEW

# Summary
- Adds validation for the redirect URL in the basic site settings.

# Review By (Date)
- Feb 26

# Urgency
- Med/Low

# Steps to Test

1. Check out this branch and clear caches
2. Navigate to `/admin/config/system/basic-site-settings`
3. Fill in a garbage url and submit
4. See validation error
5. Try a few combinations of stanford.edu domains 
6. See saved value

- D8CORE-1357
- FYI @boznik 

# See Also


stanford_profile.profile
```diff
@@ -149,3 +149,29 @@ function stanford_profile_node_access(NodeInterface $node, $op, AccountInterface
...
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state interface object.
+ */
+function stanford_profile_config_pages_stanford_basic_site_settings_form_validate(array $form, FormStateInterface $form_state) {
+  $element = $form_state->getValue('su_site_url');
+  $uri = $element['0']['uri'];
+  if (!empty($uri)) {
+    $match = preg_match('/^http(s)?:\/\/.*\.stanford.edu/i', $uri);
+    if (!$match) {
+      $form_state->setErrorByName('su_site_url', 'Only valid stanford.edu domain names allowed.');
```
pookmish: would we want to validate the url points back to this site?

sherakama: How would one do that?

pookmish: conveniently its really easy. basically it calls the url and checks an endpoint for itself. But is it something we really want?

https://git.drupalcode.org/project/domain_301_redirect/blob/8.x-1.x/src/Form/Domain301RedirectConfigForm.php#L199

sherakama: Ah, this is neat. I started down this path and was having trouble getting it to validate my local configuration URLs and it occurred to me that we may want to set this setting prior to the DNS getting pointed to a site for a launch. 

This is a dangerous setting and needs more thought on how to protect the user. I am leaving it as is and creating a new ticket to evaluate the redirect only to same site.

sherakama: https://stanfordits.atlassian.net/browse/D8CORE-1431

pookmish: I'm not sure we'd want to set the setting prior to a launch. but we can review at a later date. this seems good enough for now.

PR #119, opened 2020-02-25
D8CORE-1282: Invalidate node page cache when new menu items are added…

sherakama:
… as children.

# READY FOR REVIEW 

# Summary
- Fixes bug where you add a menu item to a nested page and the parent node cache does not clear to display the sidebar menu.

# Needed By (Date)
- Thursday

# Urgency
- Medium, will affect all users when adding their first nested page.

# Steps to Test

1. Do a fresh build of stanford_profile on the 1.x branch
2. Create a new stanford_page under the `about` page
3. Using the main navigation click on the `about` link to go to the parent page
4. See no secondary sidebar menu
5. Check out this branch
6. Create a new stanford_page under the `resources` page
7. Using the main navigation click on the `resources` link to go to the parent page
8. See the secondary sidebar navigation with your new page there

I looked into the menu_block cache and it does clear out `config:system.menu.main` tags but somewhere in the render cache that isn't enough and the parent page isn't updated. If there is a better tag or context to clear please provide some insight. 

- D8CORE

- D8CORE-1282

# See Also

### DIFF
--- a/stanford_profile.profile
+++ b/stanford_profile.profile
@@ -13,6 +13,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\menu_link_content\Entity\MenuLinkContent;
 use Drupal\node\NodeInterface;
+use Drupal\Core\Cache\Cache;
 
 /**
  * Implements hook_install_tasks().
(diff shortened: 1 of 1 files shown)


pookmish:
So this is only really the situation when a node doesn't have a sidebar menu displaying, then a child is added to it in the menu.

stanford_profile.profile
```diff
@@ -74,6 +75,22 @@ function stanford_profile_menu_link_content_presave(MenuLinkContent $entity) {
...
+  // by the menu cache tags.
+  $parent_id = $entity->getParentId();
+  if (!empty($parent_id)) {
+    list($entity_name, $uuid) = explode(':', $parent_id);
+    $menu_link_content = \Drupal::entityTypeManager()->getStorage($entity_name)->loadByProperties(['uuid' => $uuid]);
+    if (is_array($menu_link_content)) {
+      $parent_item = array_pop($menu_link_content);
+      $params = $parent_item->getUrlObject()->getRouteParameters();
+      if (isset($params['node'])) {
+        CACHE::invalidateTags(['node:'  . $params['node']]);
```
pookmish:
```suggestion
        Cache::invalidateTags(['node:' . $params['node']]);
```

sherakama:
> So this is only really the situation when a node doesn't have a sidebar menu displaying, then a child is added to it in the menu.

Aye, the menu and block work just fine if there is already the menu and block in place.

PR #120, opened 2020-02-25
D8CORE-1340 Use core link in wysiwyg for now

pookmish:
# READY FOR REVIEW

# Summary
- Use drupal cores link buttons for now so that media images can be linked.
- Rebuild node access on install.

# Need Review By (Date)
- asap

# Urgency
- high.

# Steps to Test
1. checkout this branch
1. `drush cim`
1. create a page with a wysiwyg and add an image
1. validate you can add a link to the image.

# See Also

### DIFF
--- a/src/Plugin/InstallTask/RouteRebuilder.php
+++ b/src/Plugin/InstallTask/RouteRebuilder.php
@@ -48,6 +48,7 @@ class RouteRebuilder extends InstallTaskBase implements ContainerFactoryPluginIn
    */
   public function runTask(array &$install_state) {
     $this->routeBuilder->rebuildIfNeeded();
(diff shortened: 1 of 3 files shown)


src/Plugin/InstallTask/RouteRebuilder.php
```diff
@@ -48,6 +48,7 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
    */
   public function runTask(array &$install_state) {
     $this->routeBuilder->rebuildIfNeeded();
+    node_access_rebuild();
```
sherakama: Not really related to routes but does it matter. No.

PR #126, opened 2020-02-26
D8CORE-1290: setting the width so a single banner fills the parent

jenbreese:
# READY FOR REVIEW
- I need to test this more. 

# Summary
- D8CORE-1290. A single banner was not filling the parent container if it is the only object on the page. 
- I was also able to stack images (sm/med) and have them align center, left and right. This was a noted problem by @pookmish  in the comments on this ticket. 

# Needed By (Date)
- Thursday

# Urgency
- Medium

# Steps to Test

1. Pull in this change.
2. Build.
3. Create a page with no layout and only one banner. 
4. Notice the banner fills the container. 
5. To test the other issue, add a text container with sm/med images that are aligned left, center and right. 
6. Notice they are aligned as you wanted and not a single left aligned column.
7. Try interior pages to confirm nothing funny happened with the header or the navigation bar. 
8. Make sure the header is left aligned with the column of text.

- D8CORE-1290
- Profiles since it was on the Stanford page. 

- D8CORE-1290
- @pookmish's comment in the ticket
- @kerri-augenstein - Confirmed this morning that she liked the 100% size of the banner image.

# See Also


pookmish:
I don't fully understand why the css is here vs in jumpstart_ui

jenbreese:
I put it there because I was thinking it was only for the Stanford page. But I am rethinking it all over. You think in the jumpstart UI is better?

pookmish:
Personally I think the whole section withing `.jumpstart-ui` should go in that module

sherakama:
> I don't fully understand why the css is here vs in jumpstart_ui

It is because we are targeting behaviour within the template for a specific content type rather than changing the behaviour of the template itself. The jumpart_ui layouts can and will be used for basic page, news, events, and much more and not all of them are going to have menus hide or content spacing adjustments.

pookmish:
I'm not fully onboard but i'll leave it to ya'll

1 CodeClimate issue remains

jenbreese:
Can you take a look and let me know what you think? I fixed the code climate issue. I will try to get some of Kerri's time tomorrow morning to look at it in a zoom with me. My concern is that it needs more testing.

modules/stanford_profile_styles/lib/scss/node-types/stanford_page.scss
```diff
@@ -32,6 +32,18 @@
         @include padding(null null null 80px);
       }
     }
+
+    // Main region needs to fill the whole parent in case there is not enough content to push
+    // it out.
+    .main-region {
```
sherakama: 🤔 This may belong in jumpstart_ui if we are fixing a bug on the template containers. 

https://github.com/SU-SWS/jumpstart_ui/blob/8.x-1.x/templates/layouts/three-column.html.twig#L1-L6
This block of logic may need to change to provide a `flex-lg-12-of-12` when there is no left/right content.

sherakama: @jenbreese I think this PR should take care of it: https://github.com/SU-SWS/jumpstart_ui/pull/43

jenbreese: Hi @sherakama It did resolve it.  This pull request can be closed as a not do.

PR #131, opened 2020-02-27
Created a config split for acsf modules

pookmish:
# READY FOR REVIEW

# Summary
- Move acsf modules to an acsf config split.

# Need Review By (Date)
- later time is fine.

# Urgency
- low

# Steps to Test
1. using https://github.com/SU-SWS/acsf-cardinalsites/pull/94 verify the modules are still enabled on acsf environments.

# See Also


sherakama:
So this is just another split to have an `all acsf` env?

pookmish:
yep. stanfordgryphon was trying to enable the modules that weren't there and i realized it didn't need to be.
We could even go one step further and create an `ah_other` ([any acquia](https://github.com/acquia/blt/blob/12.x/settings/config.settings.php#L77)) and move the acquia purge module into there. but i think we're ok as is for now

sherakama:
Ok, lets hold on this till next week.

pookmish:
i was thinking the same thing. i'm using this branch on gryphon now so its not a blocker.

pookmish:
should we revisit this again?

sherakama:
> Should we revisit this again?

Yeah, looks like with the Paranoia work that John is putting in we will likely need something like this. I want to take a step back this week and give it a hard think as we have several needs with upgrade paths, acsf locking, shared profiles and more, that are a part of this.

PR #145, opened 2020-03-08
D8CORE-1547 Check for node instance

pookmish:


sherakama:
I suppose I will need to write the test for this as it was my code to start with...

pookmish:
Don't spend too much time on a PHPunit test. .module files are not easy to test in those.

sherakama:
@pookmish How about a quick behat test then?

pookmish:
Works for me :+1:

PR #152, opened 2020-03-18
D8CORE-1513: Allow site managers and higher to flush caches

sherakama:
# READY FOR REVIEW 

# Summary
- Alters the permissions on the route to provide a flush cache option
- New module to do all this in prep for the move out of the profile

# Needed By (Date)
- ON HOLD.
- Review again in June

# Urgency
- None

# Steps to Test

0. Install a new site with this branch checked out or...
1. Check out this branch
2. Enable the stanford_profile_permissions module
3. Import config and clear caches
4. Log in as user 3 Morgan
5. Hover over the toolbar to see dropdown options an validate that the user can see and execute the flush cache

# See Also


sherakama:
Working on fixing the tests. Please stand by.

sherakama:
That should have been easier. 

Ready.

pookmish:
I'm not really in favor of giving users access to the clear cache. There shouldn't be any need. I know alot of our users are familiar with the way D7 worked and they dont want to let that go. Users are more likely to invalidate varnish when it isn't really necessary.
Is there a particular use case? 

I wont hold up the work here for my personal opinion.  But I wanted to raise a discussion

sherakama:
> I won't hold up the work here for my personal opinion. But I wanted to raise a discussion

I agree users won't need to do this often and would rather them not have access to it but personally, I think if they didn't have the option to at least try to clear the cache when they see something stale on their site it would cause the site_manager frustration and many nastygrams to our CESs. 

How do we decide on this? Put up a poll in #sws-team with an explanation of the pros/cons?

sherakama:
Perhaps we can start a temperature reading here:
@jbickar @rebeccahongsf @boznik @imonroe 

Would you mind chiming in on whether or not we should allow site_managers to be able to flush the cache on their site?

jbickar:
My 2¢ (and worth every penny):

Yes, it should not be necessary in D8. Yes, it will result in increased support requests. Yes, we will get frustrated later and find ourselves having this same conversation again.

I am not concerned about the minor impact to Varnish and ACSF entitlements that this implies.

pookmish:
Heres the thing in my mind. Lets say there is a scenario where varnish isn't invalided for a situation. If we add this we would probably never find the issues with the cache invalidation. Contributors adding/editing content don't have the ability to clear caches and therefore these issues will persist.
At least in the early stages of this, I would rather get those requests so that we can improve the varnish invalidation on the backend without intervention from the user.

rebeccahongsf:
+1 to John and Mike's input. People can submit help requests if they would like their site cache to be cleared. My initial concern was whether or not _Clearing the cache_ guide should continue to be published in the user guide — I went ahead to unpublished it today to prevent confusion. 😄

imonroe:
I don't have a strong opinion on this, but I'll give my impressions:
- It seems to me that `site_managers` should have the tools to manage their sites, which in my mind includes the ability to do basic troubleshooting, and clearing caches seems to be in that category.
- Mike has a good point that we may find problems which we'd otherwise miss if we prevent users from themselves correcting things which could be symptomatic of deeper issues.
- We are in the early stages of the D8 product, which means there may be hidden bugs we haven't yet seen, which also speaks to the value of encouraging maximum feedback in the near future.
- The decision now isn't necessarily set in stone for all time.  It may be that we prevent users from clearing cache for the near term, in order to maximize feedback and discover how the users are working with the sites, and what they expect to be able to do for themselves.

jbickar:
>>> The decision now isn't necessarily set in stone for all time. It may be that we prevent users from clearing cache for the near term, in order to maximize feedback and discover how the users are working with the sites, and what they expect to be able to do for themselves.

This seems like a reasonable compromise to me given the points raised above.

sherakama:
Ok, 

Let's hold off on this PR for the time being and set up a reminder to reflect on this in say, 3 months time?

sherakama:
@rebeccahongsf has there been enough/any requests to clear caches in feedback/issues from site owners that we should revisit this or can we just call this one closed?

rebeccahongsf:
@sherakama I think we can close this out! 👍  


So far there's only been one inquiry but I was able to smooth it out by explaining how cache refreshed upon saving a new node and if there were any issues that couldn't be resolved through that, then they can contact us.

sherakama:
Sounds good to me. Closing this out.

PR #176, opened 2020-04-10
D8CORE-1499: Multiple Per Row

jenbreese:
# READY FOR REVIEW
# Replacement for the Stanford Basic theme pull request. I moved it here per comments.

# Summary
- adjustments for the mulitparagraphs per row.

# Review By (Date)
- soonish

# Urgency
- high

# Steps to Test

1. Set up 4 per row: https://github.com/SU-SWS/stanford_profile/pull/159 : checkout that branch and run blt drupal:update and that'll give you the ability to have 4 per row
2. Pull in this branch.
3. Start adding content.
4. Try it out on the homepage/full width page
5. Try it on an interior page.
6. Try it on a page with a secondary navigation.

# Notes
1. I've tried it on Chrome/Mac, FF/Mac, Safari/Mac, IE, Edge.  There are some issues with FF on Mac and PC (browserstack)
2. I tried to not change any margins. I shrunk some set sizing when it overran out of the browser. In a design meeting, we thought stacking 4 across starting at LG would be helpful since it the paragraphs are so small.  

# Affected Projects or Products
- https://github.com/SU-SWS/stanford_profile/pull/159

# Associated Issues and/or People
- D8CORE-1499
- https://github.com/SU-SWS/stanford_profile/pull/159
- Any other contextual information that might be helpful (e.g., description of a bug that this PR fixes, new functionality that it adds, etc.)
- @sherakama 

# See Also
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)

PR #178, opened 2020-04-15
D8CORE-1499: Move some stuff.

sherakama:
# READY FOR REVIEW 

# Summary
- Moves a few items around 
- Removes react_paragraphs field_formatter CSS instead of overriding it with more CSS.

# Review By (Date)
- Thursday April 16th.

# Urgency
- High

# Steps to Test

1. Review code and see comments. 
2. Ask questions and go through revisions 
3. Cycle through feedback and revisions
4. Merge down to feature branch for further scrutiny 

- Multiple in-a-row
- Pull: https://github.com/SU-SWS/stanford_profile/pull/176

- D8CORE-1499

# See Also


modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
```diff
@@ -0,0 +1,144 @@
+@charset 'UTF-8';
```
sherakama, line 1: I moved this out of the WYSIWYG type and on to its own as it contains a lot more than just the WYSIWYG styles.

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
```diff
@@ -0,0 +1,144 @@
...
+
+  // Single item per row on screens less than and equal to md.
+  &.flex-container {
+    @include grid-media-max('md') {
+      display: block;
+      .paragraph-item {
+        max-width: 100%;
+      }
+    }
+  }
```
sherakama, line 110: This is a little more generic and flatter than the previous jumpstart-ui-two-column class selector but accomplished the same thing of setting the display to block for screens smaller than (md)

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// For all templates.
+.react-paragraphs-row {
+
+  // Single item per row on screens less than and equal to md.
+  &.flex-container {
+    @include grid-media-max('md') {
```
sherakama, line 104: Instead of between grid media points we can use max here.

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
```diff
@@ -0,0 +1,144 @@
...
+            max-width: 43%;
+          }
+        }
+
+        @include grid-media('lg') {
+          @include margin(null 0);
+          max-width: 50%;
+        }
+
+        @include grid-media-between('xs', 'md') {
```
sherakama, line 91: This can be `grid-media-max`

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
```diff
@@ -0,0 +1,144 @@
+@charset 'UTF-8';
+
+// For full width pages
+.jumpstart-ui--one-column {
+  .react-paragraphs-row {
```
sherakama, line 5: I remove the extra selector for the field name so we can use this on other or new fields.

modules/stanford_profile_styles/lib/scss/paragraph-types/wysiwyg/_multirow.scss
```diff
@@ -1,171 +0,0 @@
-@charset 'UTF-8';
```
sherakama, line 1: This was moved to its own place outside of the wysiwyg type.

modules/stanford_profile_styles/stanford_profile_styles.libraries.yml
```diff
@@ -41,3 +43,11 @@ stanford_profile_styles:
   css:
     theme:
       dist/css/stanford_profile_styles.css: {}
+  dependencies:
+    - stanford_profile_styles/react_paragraphs
+
+react_paragraphs:
```
sherakama, line 49: New library for these styles and added as a dependency for others.

modules/stanford_profile_styles/stanford_profile_styles.module
```diff
@@ -43,6 +43,17 @@ function stanford_profile_styles_page_attachments(array &$attachments) {
...
+/**
+ * Implements hook_library_info_alter().
+ */
+function stanford_profile_styles_library_info_alter(&$libraries, $extension) {
+  // Replace react_paragraphs field_formatter css with our own.
+  if ($extension == "react_paragraphs" && isset($libraries["field_formatter"])) {
+    unset($libraries['field_formatter']['css']['component']['js/build/css/react_paragraphs.field_formatter.css']);
+    $libraries['field_formatter']['dependencies'][] = "stanford_profile_styles/react_paragraphs";
+  }
+}
```
sherakama, line 55: This alter removes the css from the react_paragraphs module altogether and replaces it with ours. This way anywhere that module was going to add the css we new load ours and don't have to override their styles with more stronger css.

modules/stanford_profile_styles/stanford_profile_styles.module
```diff
@@ -124,17 +135,17 @@ function stanford_profile_styles_node_form_wide(array $form, FormStateInterface
...
+function stanford_profile_styles_preprocess_react_paragraphs(&$variables) {
+  // For adding grid classes to multirow paragraphs.
   foreach ($variables['rows'] as &$row) {
     foreach ($row['items'] as &$item) {
       $width = $item['width'];
       $item['attributes']->addClass("flex-$width-of-12");
     }
     $count = count($row['items']);
     $row['attributes']->addClass("container-$count-items");
+    $row['attributes']->addClass("flex-container");
```
sherakama, line 148: flex-container is needed to activate the flex-classes. This is new and wasn't required before because react_paragraphs set the container to `display:flex` but has since been removed.

modules/stanford_profile_styles/stanford_profile_styles.module
```diff
@@ -124,17 +135,17 @@ function stanford_profile_styles_node_form_wide(array $form, FormStateInterface
...
+    $row['attributes']->setAttribute('data-item-count', $count);
```
sherakama, line 149: added data-attribute to the row container as that is what data-attributes are for.

modules/stanford_profile_styles/webpack.config.js
```diff
@@ -53,6 +53,7 @@ var webpackConfig = {
     "stanford_profile_styles.node.stanford_page":         path.resolve(srcSass, "node-types/stanford_page.scss"),
     "stanford_profile_styles.paragraph.wysiwyg":          path.resolve(srcSass, "paragraph-types/wysiwyg/index.scss"),
     "stanford_profile_styles.paragraph.spacer":           path.resolve(srcSass, "paragraph-types/spacer/index.scss"),
+    "stanford_profile_styles.react_paragraphs":           path.resolve(srcSass, "paragraph-types/react_paragraphs/index.scss"),
```
sherakama, line 56: New webpack entry point for multiple row styles.

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
jenbreese: That makes sense.

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
jenbreese: I can see how this is better since there is only one variable, the 'md', so it is easier to read?

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
jenbreese: Also less specificity.

modules/stanford_profile_styles/stanford_profile_styles.module
jenbreese: That is helpful to keep us from writing stronger selectors.

jenbreese:
@sherakama I understand your changes and agree. I only found one issue with the media being too wide. I thought I'd merge and fix that once it is merged back?

jbickar:
https://www.youtube.com/watch?v=VYfpVqZSrsE&t=0m25s

modules/stanford_profile_styles/lib/scss/paragraph-types/react_paragraphs/_multirow.scss
sherakama: The 'xs' breakpoint is equal to 0 so there is nothing before it. This is a micro-optimization and in the grand scheme of things, it more helping us as developers be able to read it and prevent less hilarity if we change xs to something else later.

modules/stanford_profile_styles/stanford_profile_styles.module
sherakama: When we get into sub-theming this becomes much easier as you can do this in the theme's info.yml file.

https://www.drupal.org/docs/8/theming/adding-stylesheets-css-and-javascript-js-to-a-drupal-8-theme#override-extend

It is a super handy way of saying, "I don't want any of that, let me write my own thing and get out of the way."

PR #179, opened 2020-04-15
D8CORE-1847 fix link text getting encoded

pookmish:
# NOT READY FOR REVIEW

# Summary
- Decode html entities and quotes when the Display Suite cloned field is used.
- I have a feeling this is going to happen often.
- first commit in this PR was the test alone. following commits resolved the test.

# Need Review By (Date)
- 4/22

# Urgency
- high

# Steps to Test
1. checkout this branch
1. create a card and for the link text put some quotes in.
1. choose "Button" for the link style
1. Save and verify the quotes look correct
1. change the button style to action
1. verify the quotes still look correct

# See Also

### DIFF
--- a/modules/stanford_paragraph_card/stanford_paragraph_card.module
+++ b/modules/stanford_paragraph_card/stanford_paragraph_card.module
@@ -55,10 +55,28 @@ function stanford_paragraph_card_preprocess_ds_entity_view(array &$variables) {
     $variables['content']['#ds_configuration']['regions']['card_cta_label'] = $label;
 
     // Move the fields.
-    $variables['content']['#fields']['card_cta_label']['display_field_copy:paragraph-su_card_link_title'] = $variables['content']['display_field_copy:paragraph-su_card_link_title'];
+    $variables['content']['#fields']['card_cta_label'] = $variables['content']['#fields']['card_button_label'];
 
+    array_walk($variables['content']['#fields']['card_cta_label'], '_stanford_paragraph_card_decode_markup');
     // Unset the cruft.
     unset($variables['content']['#fields']['card_button_label']);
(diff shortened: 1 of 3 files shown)


modules/stanford_paragraph_card/stanford_paragraph_card.module
```diff
@@ -55,7 +55,7 @@ function stanford_paragraph_card_preprocess_ds_entity_view(array &$variables) {
     $variables['content']['#ds_configuration']['regions']['card_cta_label'] = $label;
 
     // Move the fields.
-    $variables['content']['#fields']['card_cta_label']['display_field_copy:paragraph-su_card_link_title'] = $variables['content']['display_field_copy:paragraph-su_card_link_title'];
+    $variables['content']['#fields']['card_cta_label'] = $variables['content']['#fields']['card_button_label'];
```
sherakama, line 58: I think this is fine but let me circle back at this a little later to grok what is happening here.

pookmish: yeah i was reading the code and it seemed weird to me that the field would be set with a variable outside the fields.

sherakama: Ok, so this works but I was getting a refresher on to what is going on here. So in order to split the link and the title into two different regions, I used a display suite field copy field to place the link field in two different regions. Then I used different formatters to get the output to what I wanted for each region (link only & title only). This line was copying the value of the copied field from one region of the template to another so that we can get an action link vs a button.  I don't see why we cannot use the original field value for the copied content. We handle the field sanitization and character encoding for the cta_label here: https://github.com/SU-SWS/jumpstart_ui/blob/8.x-1.x/templates/components/card/card.html.twig#L4 but not for the card_button label. We probably should improve on that template as well.

PR #202, opened 2020-05-16
8.1.10

pookmish:
# READY FOR REVIEW

# Summary
- Use state for the home, 404 and 403 pages since the new default content doesn't consist of node IDs.
- Added field permissions to restrict access to the site url in the site settings.

# Need Review By (Date)
- asap

# Urgency
- Critical

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-home-url-test`
1. do a fresh install `blt drupal:install`
1. Verify the home page doesn't redirect to another path like `/home`

# See Also


composer.json
```diff
@@ -130,18 +131,18 @@
...
+        "su-sws/react_paragraphs": "dev-8.x-1.x",
+        "su-sws/stanford_basic": "dev-8.x-4.x",
+        "su-sws/stanford_date_formats": "dev-8.x-1.x",
+        "su-sws/stanford_fields": "dev-8.x-1.x",
+        "su-sws/stanford_image_styles": "dev-8.x-1.x",
+        "su-sws/stanford_news": "dev-8.x-2.x",
+        "su-sws/stanford_notifications": "dev-8.x-1.x",
+        "su-sws/stanford_person": "dev-8.x-1.x",
+        "su-sws/stanford_ssp": "dev-8.x-1.x",
+        "su-sws/stanford_text_editor": "dev-8.x-1.x"
```
imonroe, line 145: We want these to be locked, right?

pookmish: oh right.. uno momento

pookmish: should be good now

PR #206, opened 2020-05-19
D8CORE-000: Set cards to 3 in a row.

sherakama:
# READY FOR REVIEW

# Summary
- Sets cards to 3-in-a-row

# Needed By (Date)
- Sooon

# Urgency
- high

# Steps to Test

1. What else needs to change? Anything?

- D8CORE

# See Also


pookmish:
What about media with caption? Or do you want to do it on the form instead of the field?

sherakama:
> What about media with caption? Or do you want to do it on the form instead of the field?

I thought media with caption was still good at 4 in a row?

pookmish:
I would think its OK with 4, but I wanted to check

PR #208, opened 2020-05-20
D8CORE-000: Set media with caption and wysiwyg to 3 in a row.

sherakama:
# READY FOR REVIEW 

# Summary
- Sets all react paragraphs to max of 3 in a row.
- Adds notification to all user types for max-of-3

# Needed By (Date)
- ASAP

# Urgency
- Today

# Steps to Test

1.  Check out this branch
2. Import config
3. Try to place items more than 3.

# See Also

### DIFF
--- a/stanford_profile.post_update.php
+++ b/stanford_profile.post_update.php
@@ -118,10 +118,17 @@ function stanford_profile_post_update_8013() {
     'contributor',
   ]);
 
-  $message = 'Update: All Paragraphs within a page will now support a maximum of 3 items per row where some allowed 4 items before. Any existing layouts with 4 items in a row will be grandfathered in but new content will be limited to 3 items per row.';
-  $notifications->addNotification($message, [
-    'site_manager',
-    'site_editor',
-    'contributor',
-  ]);
+  $database = \Drupal::database();
+  $query = $database->select('node_revision__su_page_components', 'nrspc');
+  $query->condition('nrspc.su_page_components_settings', '%"index":3%', 'LIKE');
+  $num_rows = $query->countQuery()->execute()->fetchField();
+  if ($num_rows >= 1) {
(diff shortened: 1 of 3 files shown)


stanford_profile.post_update.php
```diff
@@ -104,9 +104,24 @@ function stanford_profile_post_update_8013() {
...
-  $notifications->addNotification($message, ['site_manager', 'contributor']);
+  $message = 'New: You can now create "Person" content. See <a href="https://userguide.sites.stanford.edu/tour/person">the user guide</a> for more information';
+  $notifications->addNotification($message, [
+    'site_manager',
+    'site_editor',
+    'contributor',
+  ]);
+
+  $message = 'Update: All Paragraphs within a page will now support a maximum of 3 items per row where some allowed 4 items before. Any existing layouts with 4 items in a row will be grandfathered in but new content will be limited to 3 items per row.';
+  $notifications->addNotification($message, [
```
pookmish, line 122: I think to avoid confusion to users, we should only send the notification to those who have used 4. heres the query: `$query = 'select * from node_revision__su_page_components where su_page_components_settings like \'%"index":3%\'';`

sherakama: Thanks for the query. Have another go.

stanford_profile.post_update.php
```diff
@@ -104,9 +104,31 @@ function stanford_profile_post_update_8013() {
...
+    'site_manager',
+    'site_editor',
+    'contributor',
+  ]);
+
+  $message = 'New: You can now create "Person" content. See <a href="https://userguide.sites.stanford.edu/tour/person">the user guide</a> for more information';
+  $notifications->addNotification($message, [
+    'site_manager',
+    'site_editor',
```
pookmish, line 117: good catch on those roles

pookmish:
very nice
go ahead after tests pass

PR #210, opened 2020-05-21
D8CORE-2074: Fixing the line height in the tables to em

jenbreese:
# READY FOR REVIEW

# Summary
- changing line height to em from rem

# Needed By (Date)
- 5/26

# Urgency
- Medium

# Steps to Test

1. Pull in this branch.
2. Create a table
3. Verify multiple rows aren't overlaying each other.

- It is part of the WYSIWYG

- D8CORE-2074
- @cjwest 

# See Also


jenbreese:
Hi @pookmish Can you take a quick look?

jenbreese:
@pookmish Should I check with someone else?

jenbreese:
@kerri-augenstein Can you take a look at this pull request for possible ramifications?

modules/stanford_profile_styles/lib/scss/paragraph-types/wysiwyg/_table.scss
```diff
@@ -4,7 +4,7 @@
   table {
     th,
     td {
-      line-height: 1.2rem;
+      line-height: 1.2em;
```
sherakama, line 7: overall, we would like to stick to rems as ems get messy. Can this be bumped up to say 1.8rem or something?

kerri-augenstein: @sherakama ironically, I think line-height is where we may actually want to employ ems instead of rems, as in the case of line-height, they flex with the font size scaling across break points. 

Aka, if we use rems... 

1.8rems on MD-LG-XL, we would be forced to change 1.8rems to something different in XS-SM, like 1.6rems. 

but. 

If we use ems... 

- 1.2ems is usually always fit for the line height on headings
- 1.5ems is usually always fit for the line height on long body copy 

regardless of the font-size. 

So if ems are used for line-height, the line heights are set once, and then we forget it. 

cc @jenbreese

sherakama: I'm not going to die on this hill. We don't use ems often as they are a relative unit that when applied affects all child elements and can cause a chain of undesirable scale. As we don't encourage nesting much in tables have at it whichever way makes sense.

PR #218, opened 2020-06-05
D8 core 1431

boznik:
# READY FOR REVIEW

# Summary
- Validation that the url entered points to the site being used

# Needed By (Date)
- June 8th

# Urgency
- Low

# Steps to Test

1. Check out this branch
2. Comment out Test for stanford domain if needed on local
3. navigate to admin/config/system/basic-site-settings
4. verify you can only enter the base url of the site

- All stacks implementing this change

- https://stanfordits.atlassian.net/browse/D8CORE-1431

# See Also


stanford_profile.profile
```diff
@@ -214,6 +214,11 @@ function stanford_profile_config_pages_stanford_basic_site_settings_form_validat
     if (!$match) {
       $form_state->setErrorByName('su_site_url', t('Only valid stanford.edu domain names allowed.'));
     }
+    $host = \Drupal::request()->getSchemeAndHttpHost();
+    if ($host != $uri) {
+      $form_state->setErrorByName('su_site_url', t('Your current domain does not match the supplied domain.'));
+    }
```
sherakama, line 220: So this would force the user to be logged in on the URL they want to set the site to? I think that makes sense. @pookmish @jbickar Thoughts on this? Would this validation break the drush commands/scripts we use to set it?

@boznik can you update the help text to indicate this new restriction.

pookmish: Its a bit more restrictive than using the service from the domain_301_redirect module but it would accomplish similar validation and would be faster than the service. 
And it won't affect drush commands. we don't have any current commands to set the value, only to clear it from the database.

boznik: I did try using the service from the domain_301_redirect module, but it returned false on the checkDomain method.

pookmish: That service will only work on live publicly accessible urls. It works similarly to lets encrypt verification.

boznik: gotcha

sherakama: Thanks for the feedback. @boznik update the help text on the field and then this should be GTG.

config/sync/field.field.config_pages.stanford_basic_site_settings.su_site_url.yml
```diff
@@ -12,7 +12,7 @@ field_name: su_site_url
 entity_type: config_pages
 bundle: stanford_basic_site_settings
 label: 'Site URL'
-description: 'To redirect all traffic to a single url, input the stanford.edu url here.'
+description: 'To redirect all traffic to a single url, input the stanford.edu url here.<br><li>The domain supplied must be the same domain as the base url.</li>'
```
jbickar, line 15: I would check with Rebecca/Jacob and/or someone from the UX team on the wording here. This wording does not make sense to me, and we've already experienced people FUBARing their sites due to misunderstanding what this setting does.

stanford_profile.profile
jbickar: It would be nice to have a drush command to set it, but yes, we don't have one currently.

stanford_profile.profile
```diff
@@ -210,10 +210,17 @@ function stanford_profile_config_pages_stanford_basic_site_settings_form_validat
...
   if (!empty($uri)) {
+    // Test if the site url submitted is a *.stanford.edu domain.
     $match = preg_match('/^http(s)?:\/\/.*\.stanford.edu/i', $uri);
     if (!$match) {
       $form_state->setErrorByName('su_site_url', t('Only valid stanford.edu domain names allowed.'));
     }
+    // Test if the site url submmitted is equal to current domain.
+    $host = \Drupal::request()->getSchemeAndHttpHost();
+    if ($host != $uri) {
+      $form_state->setErrorByName('su_site_url', t('Your current domain does not match the supplied domain.'));
```
jbickar, line 221: Same here regarding the error message. Please check with Rebecca/Jacob and/or the UX team regarding wording.

config/sync/field.field.config_pages.stanford_basic_site_settings.su_site_url.yml
jbickar: Probably needs a link to the SNOW form to request a hostname.

stanford_profile.profile
```diff
@@ -255,3 +262,21 @@ function stanford_profile_xmlsitemap_link_alter(array &$link, array $context) {
 
   }
 }
+
+/**
```
boznik, line 266: @sherakama ishi wanted to get rid of the core linkfieldwidget description. what do you think of this approach?

stanford_profile.profile
pookmish: theres supposed to be one, but i couldn't get it to work. maybe i just need to do it right. :shrug: `drush config:pages-set-field-value`

stanford_profile.profile
pookmish: to me this approach seems fine if it works. kinda overkill for such a little gain. this field is normally set only 1 time and rarely touched again

stanford_profile.profile
```diff
@@ -255,3 +262,21 @@ function stanford_profile_xmlsitemap_link_alter(array &$link, array $context) {
...
+function stanford_profile_field_widget_form_alter(&$element, \Drupal\Core\Form\FormStateInterface $form_state, $context) {
+
+  // Get build info and remove core linkwidget description.
+  $build_info = $form_state->getBuildInfo();
+  if ($build_info['form_id'] = "config_pages_stanford_basic_site_settings_form") {
+    $field_definition = $context['items']
+      ->getFieldDefinition();
+
+    if ($field_definition
+        ->getType() == 'link') {
```
pookmish, line 278:
```suggestion
    if ($field_definition->getType() == 'link') {
```

stanford_profile.profile
boznik: Apparently there was a feeling that this was worth removing

sherakama: Works for me. I've been asked to remove this description from other fields throughout the accelerants as well.

boznik:
changed to not ready for review until feedback from john, rebecca and ishi is implemented.

stanford_profile.profile
```diff
@@ -255,3 +262,20 @@ function stanford_profile_xmlsitemap_link_alter(array &$link, array $context) {
...
 }
+
+/**
+ * Implements hook_field_widget_form_alter().
+ */
+function stanford_profile_field_widget_form_alter(&$element, FormStateInterface $form_state, $context) {
+
+  // Get build info and remove core linkwidget description.
+  $build_info = $form_state->getBuildInfo();
+  if ($build_info['form_id'] = "config_pages_stanford_basic_site_settings_form") {
```
boznik, line 273: cc is saying $build_info doesn't get used ???

PR #237, opened 2020-06-22
D8CORE-2205 Configure config ignore to allow for custom theme enabling

pookmish:
# READY FOR REVIEW

# Summary
- Configure config_ignore module to allow for custom themes

# Need Review By (Date)
- 6/25

# Urgency
- medium

# Steps to Test
1. use the branch in https://github.com/SU-SWS/acsf-cardinalsites/pull/162 and do a `composer install`
1. To test with config readonly enabled, comment out the Acquia environment conditions in the `global.settings.php`
1. install a fresh site or use an existing one and import config
1. create any stubbed out sub theme based on stanford_basic. you only need the `info.yml` file
1. enable the subtheme via drush `drush then [theme_name]`
1. go to the appearance page and enable the theme
1. edit the theme settings
1. run a config import and validate nothing moves with the new theme.

# See Also

### DIFF
--- a/config/sync/config_ignore.settings.yml
+++ b/config/sync/config_ignore.settings.yml
@@ -1,0 +1,9 @@
+ignored_config_entities:
+  - 'core.extension:theme'
+  - 'system.theme:default'
+  - 'block.block.*'
+  - '~block.block.seven_*'
+  - '~block.block.stanford_basic_*'
+enable_export_filtering: '0'
+_core:
+  default_config_hash: v7E8C8wJWeAW2BGohMNY1tZSBc4bexM6O62tGxecTfE
--- a/src/Config/ConfigOverrides.php
+++ b/src/Config/ConfigOverrides.php
@@ -37,9 +37,11 @@ class ConfigOverrides implements ConfigFactoryOverrideInterface {
    * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
    *   Config factory service.
    */
-  public function __construct(StateInterface $state, ConfigFactoryInterface $config_factory) {
+  public function __construct(StateInterface $state, ConfigFactoryInterface $config_factory = NULL) {
     $this->state = $state;
-    $this->configFactory = $config_factory;
+    if ($config_factory) {
+      $this->configFactory = $config_factory;
+    }
   }
 
   /**
@@ -57,7 +59,7 @@ class ConfigOverrides implements ConfigFactoryOverrideInterface {
 
     // Add to the config ignore so that it will ignore the current theme's
     // settings config.
-    if (in_array('config_ignore.settings', $names)) {
+    if (in_array('config_ignore.settings', $names) && $this->configFactory) {
       // We have to get the original values to add to the array.
       $existing_ignored = $this->configFactory->getEditable('config_ignore.settings')
         ->getOriginal('ignored_config_entities', FALSE);
--- a/src/Config/ConfigOverrides.php
+++ b/src/Config/ConfigOverrides.php
@@ -3,6 +3,7 @@
 namespace Drupal\stanford_profile\Config;
 
 use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Config\ConfigFactoryOverrideInterface;
 use Drupal\Core\Config\StorageInterface;
 use Drupal\Core\State\StateInterface;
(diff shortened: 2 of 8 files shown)


config/sync/config_ignore.settings.yml
```diff
@@ -0,0 +1,9 @@
+ignored_config_entities:
```
sherakama, line 1: uuid?

config/sync/config_ignore.settings.yml
```diff
@@ -0,0 +1,9 @@
...
+  - 'core.extension:theme'
+  - 'system.theme:default'
+  - 'block.block.*'
+  - '~block.block.seven_*'
+  - '~block.block.stanford_basic_*'
```
sherakama, line 6: We will need to remember to add the soe_subtheme to the soe_profile when we move this over.

config/sync/config_ignore.settings.yml
pookmish: Uuid is only for config entities. Not plain settings

src/Config/ConfigOverrides.php
```diff
@@ -43,6 +56,21 @@ public function loadOverrides($names) {
...
+      // We have to get the original values to add to the array.
+      $existing_ignored = $this->configFactory->getEditable('config_ignore.settings')
+        ->getOriginal('ignored_config_entities', FALSE);
+      $themes = $this->configFactory->getEditable('core.extension')
+        ->getOriginal('theme');
+      foreach (array_keys($themes) as $theme_name) {
+        $existing_ignored[] = "$theme_name.settings";
+      }
+      $overrides['config_ignore.settings']['ignored_config_entities'] = $existing_ignored;
+    }
```
sherakama, line 72: I suppose it would be possible for a site to have multiple sub-themes over time and to change from one to another. Wouldn't we want to get a list of all available themes and to ignore the settings for all of them? 

Scenario: 
* Site changes theme from stanford_basic to abc theme
* time goes on and changes theme to xyz theme which also has settings
* abc theme settings stay in DB but are not config ignored and we do a deployment
* abc theme settings get deleted

pookmish: Thats what this does. All enabled themes respective settings are ignored. Not just the currently active one. If a theme is Uninstaller thats different and will delete the settings for it anyways

tests/codeception/acceptance/SubThemeCest.php
```diff
@@ -0,0 +1,133 @@
...
+    $I->runDrush('config-import -y');
+    if (!$saml_enabled) {
+      $I->runDrush('pm-uninstall simplesamlphp_auth -y');
+    }
+  }
+
+  /**
+   * Create a stub of a subtheme based on stanford_basic.
+   */
+  protected function createTheme() {
```
sherakama, line 119: Why not just package a test theme with this profile like the unit tests do? Could save issues with systems with file permission or write issues.

tests/codeception/acceptance/SubThemeCest.php
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Run config import and adjust saml module if necessary.
+   *
+   * @param \AcceptanceTester $I
+   *   Tester.
+   * @param bool $disable_config_ignore
+   *   If config ignore module should be disabled first.
+   */
+  protected function runConfigImport(AcceptanceTester $I, $disable_config_ignore = FALSE) {
```
sherakama, line 101: I see what you are trying to do here but this could be destructive or disruptive as the codeception tests run in shuffle. This could possibly undo/unset something that was expected. I am not sure how I feel about this yet. Perhaps we can get @imonroe's big brain on this to help think through if this task could cause issues in test running.

tests/codeception/acceptance/SubThemeCest.php
pookmish: This way it creates a new theme name very test. So instead of verifying only 1 scenario, it tests a different scenario every time.

tests/codeception/acceptance/SubThemeCest.php
pookmish: all tests should assume a default state ootb. So this just resets that. If a module changes a config before this, another test should not be relying on that change

sherakama: Aye, just want a second opinion on that assumption.

imonroe: The theme creation/destruction happens all in this class, so I don't think it'll interfere with anything else; the tests may run in shuffle, but they don't run simultaneously.  I actually think this approach of spinning up a mock theme to test with is pretty smart.  It may be worth abstracting out into a mocker we can re-use later as we proceed with subthemeing and the testing associated with it.

sherakama: Thanks for the feedback

PR #242, opened 2020-06-24
D8CORE-2099: Added required fields and default image for profile.

sherakama:
# READY FOR REVIEW

See: https://github.com/SU-SWS/stanford_person/pull/94


sherakama:
Good for review again.

PR #257, opened 2020-07-15
remove view_unpublished dependency that is managed by stanford_person

pookmish:


sherakama:
To discuss in the dev chat tomorrow.

PR #261, opened 2020-07-22
D8CORE-2201: adding extlink dependency

cjwest:
# Ready for review
Ready to merge, pending
https://github.com/SU-SWS/stanford_profile_helper/pull/11

# Summary
This installs the External Link module. It's needed to let AT know when you're pointing to an external link.

# Needed By (Date)
When possible

# Urgency
Addresses accessibility issues

# Steps to Test

1. re-install your site with this branch
2. Verify that the external link <svg> appears for all external links.

D8Core

- JIRA ticket: https://stanfordits.atlassian.net/browse/D8CORE-2201
- Other PRs: https://github.com/SU-SWS/stanford_profile_helper/pull/11

# See Also


cjwest:
Thank you @sherakama, for helping me to get this up and running on my computer. 

I do like how this module works! 

I wonder, however, if we want to put it up on the stack, but not have it installed. We'll have some things to fix such as the global footer. 

Screen Shot 2020-07-23 at 7 16 30 PM

sherakama:
It does us no good to just have it up on the stack and disabled so if this gets added it should be turned on and configured properly.

I believe there is configuration that allows you to provide a css, or a few css, selectors as targets on where to apply the external link. Could you limit the external link processing to just the main body content? Also, with this, we will likely have to remove the external link icon from our events, news, and people views as well. Please start with looking into capturing the configuration for the bounding of this script to a few css selectors.

cjwest:
@sherakama ,
Sounds good!

 It seems that the `<svg>` should have the styles from `visually-hidden` applied globally, then removed as necessary.

Where would you recommend I put the styling for this?

I was thinking that it would belong at the profile level, and to put it in `stanford_profile_helper/modules/stanford_profile_styles`
From there, however, I'm not sure what file to use, or if I should create a new one.

What do you recommend?

cjwest:
See https://github.com/SU-SWS/stanford_profile_helper/pull/11

sherakama:
@cjwest I think you are looking too deep. 

I think we DO want to apply the ext link functionality to everything between the header and footer.

pookmish:
A couple things wrong with this. I put in a pr to address the incorrect config. https://github.com/SU-SWS/stanford_profile/pull/266
Also, in regards to https://github.com/SU-SWS/stanford_profile_helper/pull/11 that functionality can be accomplished within the config. see https://github.com/SU-SWS/stanford_profile/pull/266/files#diff-54fe9bb3ee6acb59450c455138275b8dR20-R22

In regards to selectors, we can define the different selectors to exclude and include. Its all options within the module config page.

cjwest:
@sherakama ,
Because this applies the `<span>` that announces "This is an external link" or similar, I'd like it to be applied to the whole page. With that we should have the icon only show in designated areas such as the WYSIWYG and not the header an footer.

cjwest:
@sherakama,

Ready for your 👀 

screenshot-localhost_32772-2020 08 13-12_19_41

sherakama:
@cjwest please review the new export meets what we discussed and glance over the codeception test.

cjwest:
@sherakama,
Thanks for your help on this. It all looks good to me.

PR #295, opened 2020-09-17
Removed unnecessary field permission third party settings

pookmish:
# READY FOR REVIEW

# Summary
- Removed field permission third party settings
- see https://github.com/SU-SWS/stanford_profile_helper/pull/26

# Need Review By (Date)
- :man_shrugging: 

# Urgency
- low

# Steps to Test
1. checkout this branch.
1. do an install or config import
1. see that it still works like normal.

# See Also

### DIFF
--- a/config/sync/field.storage.config_pages.su_event_xml_url.yml
+++ b/config/sync/field.storage.config_pages.su_event_xml_url.yml
@@ -4,7 +4,6 @@ status: true
 dependencies:
   module:
     - config_pages
-    - field_permissions
     - link
 id: config_pages.su_event_xml_url
 field_name: su_event_xml_url
--- a/config/sync/field.storage.config_pages.su_event_xml_url.yml
+++ b/config/sync/field.storage.config_pages.su_event_xml_url.yml
@@ -6,9 +6,6 @@ dependencies:
     - config_pages
     - field_permissions
     - link
-third_party_settings:
-  field_permissions:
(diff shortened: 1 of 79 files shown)


sherakama:
Won't these just keep coming back? I know we don't need them but won't exports become troublesome?

sherakama:
Oh I see https://github.com/SU-SWS/stanford_profile_helper/pull/26 now.

sherakama:
I don't really mind having these field_permissions in this profile repo but I do agree it creates a mess in our modules. The default setting is public so I think this is ok, but I want to get @imonroe's thoughts on this as well. Could there be a potential issue with removing these? IE: Change the default setting and have all fields go away all of a sudden?

pookmish:
Removing the third party settings basically results in the same outcome as the `public` settings. The tests would have caught any issues with fields disappearing.

This PR is a result of implementing https://github.com/SU-SWS/stanford_profile_helper/pull/26. After a config import occurs (site install), it saves all the field storage entities which then runs through the hook in the profile helper. The profile helper removes the third party setting and then I exported the config. This is a result of that.

imonroe:
I can't think of any immediate objections to this, but it also seems to me like just a little housekeeping.  I'm not sure we're getting any benefit besides parsimony, but that's probably enough to make it worthwhile.  I say thumbs-up.

imonroe (approved):

PR #298, opened 2020-09-21
D8CORE-1200 Additional permission check for deleting home page

pookmish:
# READY FOR REVIEW

# Summary
- I built out this test, so we might as well use it. It verifies the user can't bulk delete the home page.

# Need Review By (Date)
- 10/5 i guess

# Urgency
- low

# Steps to Test
1. View the tests.

# See Also

### DIFF
--- a/tests/codeception/functional/Users/RolesCest.php
+++ b/tests/codeception/functional/Users/RolesCest.php
@@ -48,11 +48,17 @@ class RolesCest {
 
   /**
    * D8CORE-1200 Prevent deleteing the homepage from bulk delete.
+   *
+   * @group testme
    */
   public function testBulkDeleteHomePage(FunctionalTester $I) {
     $test_home = $I->createEntity([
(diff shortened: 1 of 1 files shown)


sherakama (approved):
GTG on green lights.

pookmish:
yeah to get full green lights might take awhile since it's related to some deep underlying dependency

sherakama:
> yeah to get full green lights might take awhile since it's related to some deep underlying dependency

Could be related.
https://www.drupal.org/project/upgrade_status/issues/3172382

pookmish:
That issue seems to be the same. But it's not specific to upgrade_status. see https://github.com/mglaman/drupal-check/issues/186#issuecomment-696856219
I was able to use `phpstan/phpstan:0.12.42` and no issues. maybe we lock caravan to that version for now? what do you think?

pookmish:
Cool. green lights now

PR #310, opened 2020-10-14
D8CORE-2856 Add list and entity reference paragraph types

pookmish:
# READY FOR REVIEW

# Summary
- Created entity reference and view reference paragraph types
- Stubbed out starting point for a "Card" display for each paragraph type. These will be revised later
- Limited the views to only a few choices that I assume are the best at this time.

# Need Review By (Date)
- 10/19

# Urgency
- medium

# Steps to Test
1. checkout this branch and the same branch in https://github.com/SU-SWS/stanford_profile_helper/pull/34
1. `composer require drupal/viewfield` (it's easier this way than doing all the different requirements)
1. `blt drupal:update` or `blt drupal:install -n`
1. Create a basic page and use the "Lists" or "Content Reference" paragraph types.
1. verify they work close to what you'd expect.

# See Also

see https://github.com/SU-SWS/stanford_profile_helper/pull/34

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -138,7 +138,7 @@
         "su-sws/stanford_news": "dev-8.x-2.x",
         "su-sws/stanford_notifications": "dev-8.x-1.x",
         "su-sws/stanford_person": "dev-8.x-1.x",
-        "su-sws/stanford_profile_helper": "dev-8.x-1.x",
+        "su-sws/stanford_profile_helper": "dev-D8CORE-2856 as dev-8.x-1.x",
         "su-sws/stanford_ssp": "dev-8.x-1.x",
         "su-sws/stanford_text_editor": "dev-8.x-1.x"
     },
--- a/composer.json
+++ b/composer.json
@@ -122,6 +122,7 @@
         "drupal/role_delegation": "^1.0@beta",
         "drupal/seckit": "^1.0@alpha",
         "drupal/ui_patterns": "~1.0",
+        "drupal/viewfield": "^3.0@beta",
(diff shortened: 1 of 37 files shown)


sherakama:
Running an install now but the tests are passing to this is likely fine.

Do you want to move the new view modes down into the child projets? Eg: the news and events card views down into the news/events modules?

pookmish:
`:meh:` :shrug: I'm indifferent about moving them down into the modules. I'd say at least at this point let's wait. I'm sure some tweaks will come to the card displays. at that time we can decide if it's worth that effort.

sherakama:
> `:meh:` 🤷 I'm indifferent about moving them down into the modules. I'd say at least at this point let's wait. I'm sure some tweaks will come to the card displays. at that time we can decide if it's worth that effort.

Ticket!

pookmish:
https://stanfordits.atlassian.net/browse/D8CORE-2869 for future sprints

sherakama:
Screen Shot 2020-10-16 at 10 16 19 AM

sherakama:
Probably should add some no content text to the views.

Also, do we want to only allow 1 per row? They get kinda squishy when put side by side.

pookmish:
> Probably should add some no content text to the views.

I was thinking we could either do that and also have a behavior checkbox that says "Hide all fields if the list is empty". It would end up being an empty space but maybe we can help with that. :shrug:

sherakama (approved):
Ok, sounds like you got this covered. I will mark as approved and let you merge when you are good and ready.

PR #337, opened 2020-11-24
D8CORE-000 Patch simplesamlphp auth module to prevent unwanted redirect

pookmish:
# READY FOR REVIEW

# Summary
- Fixes the redirect the next day.

# Need Review By (Date)
- 12/3

# Urgency
- medium

# Steps to Test
1. apply the patch manually or `composer require su-sws/stanford_profile:dev-D8CORE-000-saml-patch`
1. log into your local with saml.
1. navigate to a page like `/events`
1. see you are still logged in and on that url still.
1. edit the file `docroot/modules/contrib/simplesamlphp_auth/src/EventSubscriber/SimplesamlSubscriber.php` and comment out part of the code as in the screenshot below.
1. reload your `/events` page
1. verify you stayed on that page and are logged out.
image

# See Also

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -178,8 +178,8 @@
             "drupal/paragraphs": {
                 "https://www.drupal.org/project/paragraphs/issues/2901390": "https://www.drupal.org/files/issues/2020-06-25/paragraphs-2901390-51.patch"
             },
-            "drupal/views_taxonomy_term_name_depth": {
-                "https://www.drupal.org/project/views_taxonomy_term_name_depth/issues/2877249": "https://www.drupal.org/files/issues/2020-05-01/2877249_21.patch"
+            "drupal/simplesamlphp_auth": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -178,8 +178,8 @@
             "drupal/paragraphs": {
                 "https://www.drupal.org/project/paragraphs/issues/2901390": "https://www.drupal.org/files/issues/2020-06-25/paragraphs-2901390-51.patch"
             },
-            "drupal/views_taxonomy_term_name_depth": {
```
sherakama, line 181: Just to confirm, the `views_taxonomy_term_name_depth` went away right?

The simplesamlphp patch looks good.

pookmish: The module is still there... the patch is inherited by the dependencies.

sherakama:
Should we add a notification to the user? "User was logged out due to lack of activity"

pookmish:
problem with that would be the notification could then get cached by varnish.

sherakama:
Aye, could be troublesome. Nice to have.

imonroe (approved):
Performs as described.  G2G.

PR #340, opened 2020-12-03
Exported configs after D9.1 database updates

pookmish:
# READY FOR REVIEW

# Summary
- Due to the theme dependency changes, we need to move to exclusively D9 with this.
- Ran an install with the latest D8, install D9, did a database update and config export.
- Updated version of modules for D9.
- Do config overrides to manipulate the config_ignore settings when doing a site install. We need this because the system.theme configs would be ignored during install.

# Need Review By (Date)
- 1/15/2021

# Urgency
- medium

# Steps to Test
1. follow steps in https://github.com/SU-SWS/acsf-cardinalsites/pull/212

# See Also


sherakama (changes requested):
Changes look legit. Passing tests is your merge blocker.

sherakama:
Ok, @imonroe please have a pass at this before merge.

imonroe (approved):
Looks good.  What's up with CodeClimate on this?  Looks like it just hung?

pookmish:
Codeclimate will only scan when the merge is going into the "Default" branch. right now the default branch is set to `8.x-1.x`. So just before I merge this i'll switch default branches and trigger Codeclimate.

sherakama:
> Codeclimate will only scan when the merge is going into the "Default" branch. right now the default branch is set to `8.x-1.x`. So just before I merge this i'll switch default branches and trigger Codeclimate.

This bugs me very much about code climate. If anyone has some time perhaps look into https://www.codacy.com/ as an alternative. I am not married to codeclimate. It is ok but could be better.

PR #344, opened 2020-12-08
Dont remove special characters in person names

pookmish:
# READY FOR REVIEW

# Summary
- Hotfix to prevent special characters from being removed in person names.

# Need Review By (Date)
- asap

# Urgency
- high.

# Steps to Test
1. checkout this branch
1. `drush config-import -y`
1. create a person content and use special characters like dash or & in the first or last name
1. save and verify the special characters are kept.

# See Also


imonroe (approved):
Looks good, tests green.

config/sync/auto_entitylabel.settings.node.stanford_person.yml
```diff
@@ -1,6 +1,6 @@
 status: 1
 pattern: '[node:su_person_first_name] [node:su_person_last_name]'
-escape: true
+escape: false
```
jbickar, line 3: What are the security implications of this change?

pookmish: No different than putting special characters in the title on other content types. shouldn't be any security issue.

pookmish: this is also the same settings we currently have for news content types.

jbickar: So Drupal sanitizes it prior to putting it in the DB? Can you add a [Little Bobby Tables](https://xkcd.com/327/) test?

jbickar:
You purposefully titled this PR about removing special characters `Dont remove special characters in person names` to give me the twitches, didn't you?

config/sync/auto_entitylabel.settings.node.stanford_person.yml
sherakama: Drupal doesn't often sanitize writing to the db, but rather, wants you to sanitize on output.

jbickar: That was my recollection, but I did not phrase my question articulately. Is it safe to say that the DB abstraction layer protects against this not being escaped?

PR #359, opened 2021-02-17
Moved patches to shared package and use Stable 9 theme.

pookmish:
# READY FOR REVIEW

# Summary
- Moved patches to the shared patches package.
- Use stable 9 theme and the latest stanford_basic theme.

# Need Review By (Date)
- :shrug: 

# Urgency
- low

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-drupal-patches`
2. `drush cim -y`
3. view the site and ensure you don't get any errors

# See Also

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -96,6 +96,7 @@
         }
     ],
     "require": {
+        "ext-imagick": "*",
         "composer/installers": "^1.2",
         "drupal/address": "^1.2",
         "drupal/admin_toolbar": "^2.2",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -146,8 +147,8 @@
         "drupal/views_block_filter_block": "^1.0@beta",
         "drupal/views_bulk_operations": "^3.6",
         "drupal/xmlsitemap": "^1.0",
-        "ext-imagick": "*",
         "onlyextart/colorbox": "dev-master#e58476becbc89dc671093d1bcd9f99b2167fa8f7",
+        "su-sws/drupal-patches": "^8.0",
```
sherakama, line 151: Cool. It will be interesting if/how this gets versioned when patches get rolled and our different teams keep up at different paces.

Any thought to ditching patches altogether and instead, running forks?

pookmish: I think forks is going to be really tough for those instances that we need 2 or more patches. And patches keep the rest of the module up to date so it's tough to argue in favor of forks at this time.

imonroe (approved):
Ok by me.  I added `drupal-patches` to the list of modules to check when we do releases in hal9k.

PR #360, opened 2021-02-17
D8CORE-3183: Corrected Events view display for filtered list page

imonroe:
# READY FOR REVIEW

# Summary
- Ticket D8CORE-3183 calls for some spacing between date strings in the event list view. Turns out, the problem was occurring because the `List Page - Filtered` display was different in a few ways from the `List` display.  To correct this, we update the `List Page - Filtered` display to match the `List` display, and then ensure our contextual filter works by taxonomy term id, same as before.

# Needed By (Date)
- 3/1
- 
# Urgency
- non-urgent

# Steps to Test

1. Pull this branch.
2. Do a `drush cim -y`
3. Make a multi-day event.
4. Verify that the display on the `/events` page is the same as when you click into the view filtered by event type.

### DIFF
--- a/config/sync/views.view.stanford_events.yml
+++ b/config/sync/views.view.stanford_events.yml
@@ -2350,58 +2350,9 @@ display:
     position: 2
     display_options:
       display_extenders: {  }
-      display_description: ''
-      block_description: 'List Page - Filtered'
-      css_class: stanford-events-list-page--filtered
-      defaults:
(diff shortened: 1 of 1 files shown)


pookmish:
I believe this is now resolved from another PR. can you confirm @imonroe

imonroe:
Can confirm.

image

PR #395, opened 2021-04-01
D8CORE-2572 Added and configured content_lock to prevent simultaneous edits

pookmish:
# READY FOR REVIEW

# Summary
- Added and configured [content_lock](https://www.drupal.org/project/content_lock) module

# Need Review By (Date)
- 4/7

# Urgency
- low

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-D8CORE-2572`
2. `blt drupal:update` or `blt drupal:install`
3. `drush pmu simplesamlphp_auth`
4. log in as a user and go to edit a node
5. in another browser or in an incognito window, log in as another user and edit the same node
6. verify you get a message and you can't save the form.

# See Also


config/sync/user.role.contributor.yml
```diff
@@ -19,6 +19,7 @@ permissions:
   - 'access video_browser entity browser pages'
   - 'add citation entities'
   - 'administer main menu items'
+  - 'break content lock'
```
imonroe, line 22: are we sure we want to let contributors break the locks?  Maybe we ought to reserve that for higher level roles only?  I'm not married to the idea, but it's something worth considering.

imonroe (commented):
Why remove the test?

config/sync/user.role.contributor.yml
pookmish: There's not really much concern in my mind. It's not really dangerous it's just annoying to users when they simultaneously edit which is rare in the first place.

pookmish:
I couldn't produce a test because once a user logs out, all of their locks are removed. So it's not possible to automatically test the functionality unfortunately.

PR #400, opened 2021-04-13
8.2.5

imonroe:
# READY FOR REVIEW

# Summary
- Some small tweaks to basic page plus functionality from feedback from Ishi.

# Needed By (Date)
- before 2.1.0 deployment.

# Urgency
- medium

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


config/sync/core.entity_form_display.node.stanford_page.default.yml
```diff
@@ -90,14 +89,10 @@ content:
     region: content
     third_party_settings: {  }
   su_basic_page_type:
-    type: options_shs
+    type: options_select
```
pookmish, line 92: Why this change?

config/sync/layout_library.layout.stanford_basic_page_full.yml
```diff
@@ -59,6 +59,36 @@ layout:
...
+      extra_classes: ''
+      centered: centered-container
+      columns: default
+      label: 'page description'
+    components:
+      5659aa07-22e4-461e-b826-00d505ba2d0c:
+        uuid: 5659aa07-22e4-461e-b826-00d505ba2d0c
+        region: main
+        configuration:
+          id: 'field_block:node:stanford_page:su_page_description'
```
pookmish, line 74: Ooh we missed that on the full width page.

pookmish (commented):
Should we just commit this to the dev branch incase any additional hotfixes come through?

config/sync/views.view.stanford_basic_pages.yml
```diff
@@ -572,7 +572,7 @@ display:
           relationship: none
           group_type: group
           admin_label: ''
-          default_action: ignore
+          default_action: empty
```
pookmish, line 575: I know the reasoning for this, but it also is confusing to a user that doesn't input any argument into the paragraph data field. So there's a trade off with this for the content author

config/sync/core.entity_form_display.node.stanford_page.default.yml
imonroe: Ishi found the simple hierarchal select was confusing, in that it appeared to the inexperienced user to apply two or more terms if you choose a term below the root level of the taxonomy.  By changing it to a simple select, it's more obvious that only one term is being applied.

config/sync/layout_library.layout.stanford_basic_page_full.yml
imonroe: Yeah, it was tricky to remember that the layout isn't the same between default and full-width, but I think I got it right here.

config/sync/core.entity_form_display.node.stanford_page.default.yml
pookmish: i guess that kinda makes sense but could also be resolved with help text. nothing i have a strong opinion about though.

config/sync/views.view.stanford_basic_pages.yml
imonroe: Yeah, this is another thing Ishi and I went over in a call.  It doesn't make sense to show all basic pages, and since you can use multiple terms, you can construct the list how you like.  It's a documentation thing to make sure the users know how to add the filter.

config/sync/layout_library.layout.stanford_basic_page_full.yml
pookmish: Did it have to be a whole other layout? could it just be added to the same section as the page title? it's just a little less confusing when editing.

config/sync/views.view.stanford_basic_pages.yml
pookmish: If we get user's commenting on it, we can always adjust as necessary.

config/sync/layout_library.layout.stanford_basic_page_full.yml
imonroe: I made it match the way it was set up in the default layout, and they both use the same theming. Do you mean confusing when editing the layout, or confusing for content authors?

pookmish: Yeah the result is the same. i don't really care in the long run. 1 layout with 2 components vs 2 layouts with 1 component each. it's still the same outcome.

pookmish: So let's remove this and let's also remove the description from the regular node display.

imonroe: This is fixed.  Page description removed from both the default and full page layout.

pookmish (approved):

PR #464, opened 2021-11-10
D8CORE-4246 Add fontawesome module for wysiwyg icon support

pookmish:
# READY FOR REVIEW

# Summary
- Add the fontawesome module so that the icons display in the wysiwyg

# Need Review By (Date)
- 11/19

# Urgency
- medium

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-D8CORE-4246`
2. `blt drupal:update` or `blt drupal:install`
3. create a page with a text area
4. fill the text area with some text like `<p class="fab fa-drupal">This is some text</p>`
5. view the ckeditor in normal style and verify the icon displays while editing.

# See Also

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -123,6 +123,7 @@
         "drupal/field_group": "^3.0@rc",
         "drupal/field_permissions": "^1.0@RC",
         "drupal/field_validation": "^1.0-alpha",
+        "drupal/fontawesome": "^2.19",
         "drupal/google_analytics": "^3.1",
         "drupal/google_tag": "^1.4",
         "drupal/imagemagick": "^3.1",
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -62,6 +62,7 @@ module:
   file_mdm_font: 0
   filter: 0
   focal_point: 0
+  fontawesome: 0
   google_analytics: 0
   google_tag: 0
   hal: 0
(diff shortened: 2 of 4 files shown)


jenbreese:
Is there another step to be done? I did the ones in this PR and the ones in https://github.com/SU-SWS/stanford_basic/pull/256. I'm not getting what I think you want. 
This is what I am seeing.
Screen Shot 2021-11-10 at 3 54 35 PM

pookmish:
> Is there another step to be done? I did the ones in this PR and the ones in [SU-SWS/stanford_basic#256](https://github.com/SU-SWS/stanford_basic/pull/256). I'm not getting what I think you want. This is what I am seeing. Screen Shot 2021-11-10 at 3 54 35 PM

You might have to clear your caches to accept the font styles. the word `Stuff` should have the same font styling as the other text on that page.

pookmish:
Oh thanks @jenbreese i see that i have to do the inherit on the `font-weight` as well.

jenbreese:
The weight problem is fixed.  It is a little weird to use. 
Here is the Source view.
Screen Shot 2021-11-10 at 4 49 29 PM

If you press return, you don't get a line return,  you get another icon. Maybe solvable with documentation?
Screen Shot 2021-11-10 at 4 49 15 PM
Screen Shot 2021-11-10 at 4 47 24 PM

pookmish:
Yeah the issue you are seeing appears to just be a CKEditor "Feature". Because the FA classes are a more advanced thing, I'm not too concerned about it existing, but we can document it.

jenbreese:
Great. I'll merge it.

jenbreese (approved):
GTG

PR #511, opened 2022-03-21
D8CORE-3975 Created shared tags vocabulary and fields

pookmish:
# READY FOR REVIEW

I still have to build out tests, but please review if you can. Just don't merge until I have tests added.

# Summary
- Created "Shared Tags" vocab
- Created a taxonomy field on all content types for the "Shared Tags"
  - Hidden from all displays
- "Shared Tags" view with card display using css grid from [stanford_profile PR](https://github.com/SU-SWS/stanford_profile_helper/pull/187)
- Renamed taxonomy fields
- Deleted page manager pages in favor of layout builder on term pages.
   - This will allow customizing the individual term pages if they so choose.

# Need Review By (Date)
- 4/20

# Urgency
- medium

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-shared-tags`
2. checkout the matching branches in [stanford_profile PR](https://github.com/SU-SWS/stanford_profile_helper/pull/187) & [stanford_events PR](https://github.com/SU-SWS/stanford_events/pull/107)
3. Update/install a site using one of the options below:
   - Install clean `blt di`
   - (Recommended) Update an existing `blt drupal:update`
      - I recommend using `amptesting` site to test with. It's got alot of content already for easy reference. `drush sql-sync @amptesting.01live @self; blt drupal:update; blt drupal:update`
4. View the taxonomy term pages to verify they look nearly the same as before (this is why syncing to amptesting is recommended)
5. Create some content of various content types, using some "Shared Tags"
6. Create a basic page and use the "Shared Tags" view in the list paragraph
   - I've configured multiple contextual arguments. for the arguments field you can put the taxonomy name only. This will produce a list of all content with that term. Or you can do something like `term-name/stanford_events+stanford_news` to get only events and news tagged with that taxonomy term.

# See Also


imonroe:
I also like that this is almost entirely configs, with a few tests.  Low code = good code, IMO.

pookmish:
Interesting find when there are no terms available. I've created an [issue with the module](https://www.drupal.org/project/shs/issues/32725880).

pookmish:
While looking at the issue you mentioned, I realized there's no way to "Deselect" a value that is saved. I'll try to fix that at the same time.

pookmish:
@imonroe just need a PR approval and this will merge in.

imonroe (approved):
LGTM

PR #522, opened 2022-04-14
localist terms

pookmish:
# READY FOR REVIEW

# Summary
- Add 3 taxonomy vocabs and fields on events
- Import the content from Localist for the vocabs
- Configure the event list view to take those vocabs and additional arguments
- Set the overwrite_properties on importers to allow unpublishing content. (exposes the field on the form)

# Need Review By (Date)
- 4/22

# Urgency
- Medium

# Steps to Test
1. checkout this branch
2. import configs
3. import events from localist
4. Use the "List" paragraph type to produce a list of events
5. use the arguments field to build a list of events in the form `audiences+types/groups+departments/subjects/keyword+tags`
    - To skip an argument you can provide empty quotes: `''/''/music` 
    
# See Also


imonroe (approved):
This works as described.  Mostly configs, with a tests that pass.  I can see now why you found that chunking/sub_process workaround in migrations.

PR #526, opened 2022-04-19
D8CORE-5729 People term pages: display only child terms groupings

pookmish:
# READY FOR REVIEW

# Summary
- If a person is tagged with multiple types/groups, we don't want to display types/groups that are not relevant to the current taxonomy term we are viewing.
- This actually also fixes the ticket D8CORE-5657 which needed the headers of the lists adjusted.

# Need Review By (Date)
- 4/28

# Urgency
- medium

# Steps to Test
1. checkout this branch
2. import configs
3. configure the taxonomy terms so that there are at least 2 parents with some child tags under those.
4. create a person and tag with multiple child terms from different parents
5. view the parent taxonomy term page
6. verify the child from the other parent isn't displayed.

# See Also

### DIFF
--- a/config/sync/core.entity_view_display.taxonomy_term.stanford_person_types.default.yml
+++ b/config/sync/core.entity_view_display.taxonomy_term.stanford_person_types.default.yml
@@ -74,19 +74,6 @@ third_party_settings:
               expand_all_items: false
             weight: 0
             additional: {  }
-          a26cc41e-5521-4f30-8180-936d156bae3c:
-            uuid: a26cc41e-5521-4f30-8180-936d156bae3c
-            region: main
-            configuration:
-              id: 'views_block:taxonomy_term_pages-results_for'
-              label: ''
-              label_display: '0'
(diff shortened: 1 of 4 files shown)


jenbreese:
I think this works. I tagged new content like this:
Screen Shot 2022-04-19 at 3 29 22 PM
And in the menu, only the first child is displayed.
Screen Shot 2022-04-19 at 3 30 19 PM

pookmish:
It's more about the list grouping. If you are on the "Family" term page, you shouldn't see the grouping for "UIT" or "Lecturer". Similarly, if you are on the "Staff" term page, you shouldn't see "Uncle" or "Lecturer".

These used to be headings in above the people content.

jenbreese:
Thank you for the clarification. It works like you said in the last comment. I don't see the grouping for "UIT" or "Lecturer" on the Family page. On the "Staff" term page, I don't see "Uncle" or "Lecturer".

PR #557, opened 2022-05-18
D8CORE-3715 Updated field encrypt to 3.0.0

pookmish:
# DO NOT MERGE

# Summary
- Updated field encrypt module to 3.0.0

# Need Review By (Date)
- :shrug: 

# Urgency
- low

# Steps to Test
1. using the 8.x or 9.x branches, add some values to the person importer credentials field.
2. checkout just the update file from this PR: `git checkout D8CORE-3715 -- stanford_profile.post_update.php`
   - What we will probably do is run the functions via `drush eval` just before doing a code deployment because we have to decrypt and uninstall the module before we update it.
4. `drush updb`
5. require this branch `composer require su-sws/stanford_profile:dev-D8CORE-3715`
6. `blt drupal:update` or `drush deploy`
7. view the person importer config page
8. verify your values are still as expected.

# See Also


imonroe:
I think in the testing steps, we need to run `drush cron` after doing the post_update 
image

pookmish:
right.. i did put that into the jira notes. We can manually run cron or just let normal scheduled cron do the task.

pookmish:
So the cron is only required at the very end. If you see in the database updates part it decrypts the values.
at the end you are left with the unencrypted values. but then cron runs and it encrypts them.

imonroe (approved):
This worked correctly for me as described.

PR #574, opened 2022-06-13
Improve event sponsor field update hook

pookmish:
# READY FOR REVIEW

# Summary
- Fix update hook. It breaks currently on ACSF deployment because the configuration in the database doesn't match the file config and so it breaks when trying to import configs.
- see https://evolvingweb.ca/blog/resizing-fields-drupal-8-without-losing-data

# Need Review By (Date)
- asap

# Urgency
- 

# Steps to Test

# See Also

### DIFF
--- a/modules/stanford_events/stanford_events.install
+++ b/modules/stanford_events/stanford_events.install
@@ -5,6 +5,8 @@
  * Stanford_events.install.
  */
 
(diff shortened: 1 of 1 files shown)


imonroe (approved):
Great catch.  Thank you for pointing this out to me.  I'll be sure to use this pattern for future update hooks.

PR #617, opened 2022-09-20
D8CORE-6288 Policy content type

pookmish:
# READY FOR REVIEW

# Summary
- Policy content type
- Works with https://github.com/SU-SWS/stanford_fields/pull/35

# Need Review By (Date)
- ASAP

# Urgency
- high

# Steps to Test
1. `composer require su-sws/stanford_profile:dev-D8CORE-6288`
2. install new site or import configs
3. create a new "Policy" node
   1. The first node will be the "Book" that will bind all the policies togeather
   2. In the right sidebar, choose `- Create a new book -`
   3. Populate other fields as desired
   4. Save the node
4. Repeat the above step but choose the "Book" you created above.
5. Choose it's parent item and it's position similar to adding a page to the main menu.
6. Repeat with various parent levels and different orders
7. Verify the left sub navigation changes and matches what is expected on each save.
8. Populate some of the change logs and verify they show on the bottom of the page
9. Go to the config page at `/admin/config/content/policy`
10. change the prefix settings
11. review the changes on the node pages to verify your choices have been reflected.


pookmish:
I'm working on the tests and fixing some stuff with the prefixing titles. but it can be reviewed for other parts for now.

pookmish:
I believe I have this in a good situation now. Finally got the prefixing to work as I wanted and all the tests pass. @jenbreese mentioned to me in Slack that it's working for her as well.

imonroe:
If you create a new policy without creating a book, it will create it, but it doesn't create it properly: 
image

If you try to create the book after doing so, it will not let you change the label of the book, nor will it let you change the Related Policies value.

I deleted my broken policy book from above, and tried to recreate it using the Book settings, and adding an initial changelog entry, and I got an integrity constraint violation for the primary key: 

image

Should the default behavior be to perhaps automatically create a book if a new Policy is created but no Books exist?

I'm resetting my db to baseline and continuing to test.

pookmish:
@imonroe delete `docroot/modules/contrib/*` and re-run `composer install`. some patches didn't get applied.

imonroe:
Nixing the `contrib` dir did the trick for me Mike.  

Functions as I would expect at this point.  I think it needs a little UI tweaking for the breadcrumbs, dates and prev/next buttons perhaps? Like spacing and styling, but that's ok.

There's also no logic in terms of "effective date" or "last updated".  That seems ok, but it also seems like the effective date should always be on or after the last updated?  

The printable version works for the copy but I get a broken image in the header.  Super good that it has the source URL in the footer.

jenbreese:
@imonroe I see the broken image on the printable page too.  It is trying to load this: ``

jenbreese (approved):
It works well for me. I was able to move pages around in the book and can use the comment logs.  I can see all the pieces and can get to them to style.

imonroe:
Jen would like to continue on this, so I'm going to go ahead and merge it.

PR #701, opened 2023-08-21
D8CORE-6842 Added and configured stanford_samlauth

pookmish:
# READY FOR REVIEW

# Summary
- Implement Saml auth module and configurations

# Review By (Date)
- ASAP

# Review Tasks

1. checkout the branch in https://github.com/SU-SWS/acsf-cardinalsites/pull/305
2. do a fresh install or update an existing site
3. log into the site
4. verify you get the expected roles and message.


sherakama:
@pookmish anyway we can share this review with more developers? This would be good for our https://stanfordits.atlassian.net/browse/OKRSPY24-7 OKR.

pookmish:
@sherakama All devs, including you, were tagged in https://github.com/SU-SWS/stanford_samlauth/pull/1
Only 2 responded. all are welcome to review that PR still.

imonroe (approved):
Tested with https://github.com/SU-SWS/acsf-cardinalsites/pull/305 and it works for me.

There is one tiny thing I noticed that we should probably fix.  When a new site is spun up, on the homepage, in the top banner "Welcome to your site" block, there is a login link.  The link works, and sends you through the saml process correctly, but when it redirects back, it redirects to `/saml_login`, and the user is presented with an "Access Denied" page.  

The "Web Login" button in the footer has a better link, which is set to `/saml/login?destination=/`.  I suggest we alter the link in the "welcome to your site" banner at the top to match, to avoid confusion by users spinning up a new site.

pookmish:
ahh good catch. Luckily that's just default content we can update

PR #855, opened 2025-01-28
D8CORE-7622: Site Reviewer Role

buttonwillowsix:
# READY FOR REVIEW

# Summary
- Adding a reviewer role that allows reviewers to see unpublished pages without seeing the editing interface.

# Review By (Date)
- Jan 30.

# Criticality
- Medium

# Urgency
- Normal

# Review Tasks

1. Check out this branch
2. Make an unpublished page
3. Change role to Reviewer
4. Visit the unpublished page
5. Hopefully see it
6. Note that you can log out an that there is an indication that the page is unpublished

### Site Configuration Sync

- Is there a config:export in this PR that changes the config sync directory? YES

## Front End Validation
No front-end impact

### Code
- [ YES] Are the naming conventions following our standards?
- [YES ] Does the code have sufficient inline comments?
- [NO ] Is there anything in this code that would be hidden or hard to discover through the UI?
- [ NO] Are there any [code smells](https://blog.codinghorror.com/code-smells/)?
- [Sort of? ] Are tests provided? eg (unit, behat, or codeception)

### Code security
- [ N/A] Are all [forms properly sanitized](https://www.drupal.org/docs/8/security/drupal-8-sanitizing-output)?
- [NO ] Any obvious [security flaws or new areas for attack](https://www.drupal.org/docs/8/security)?

## General
- [NO ] Is there anything included in this PR that is not related to the problem it is trying to solve?
- [ YES] Is the approach to the problem appropriate?

- Stanford Sites

- D8CORE-7622

### DIFF
--- a/config/sync/user.role.site_reviewer.yml
+++ b/config/sync/user.role.site_reviewer.yml
@@ -7,5 +7,4 @@ label: 'Site Reviewer'
 weight: 4
 is_admin: null
 permissions:
-  - 'access toolbar'
   - 'view any unpublished content'
--- a/config/sync/user.role.site_reviewer.yml
+++ b/config/sync/user.role.site_reviewer.yml
@@ -6,4 +6,6 @@ id: site_reviewer
 label: 'Site Reviewer'
 weight: 4
 is_admin: null
-permissions: {  }
+permissions:
+  - 'access toolbar'
+  - 'view any unpublished content'
--- a/config/sync/user.role.site_reviewer.yml
+++ b/config/sync/user.role.site_reviewer.yml
@@ -1,0 +1,9 @@
+uuid: 789e9ccb-b42b-44e2-9dc4-dea61e0c4ac4
+langcode: en
+status: true
+dependencies: {  }
+id: site_reviewer
+label: 'Site Reviewer'
+weight: 4
+is_admin: null
+permissions: {  }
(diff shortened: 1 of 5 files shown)


config/sync/user.role.site_reviewer.yml
```diff
@@ -0,0 +1,11 @@
...
+langcode: en
+status: true
+dependencies: {  }
+id: site_reviewer
+label: 'Site Reviewer'
+weight: 4
+is_admin: null
+permissions:
+  - 'access toolbar'
+  - 'view any unpublished content'
```
pookmish: I don't think they need access to the toolbar do they? They won't have any buttons to click.

buttonwillowsix: I will remove that!

As discussed, I added this on the SDSS stack so they would have an option to logout. Mike and I said we would  keep this more slim by removing it and add it later as needed. I think we can also document how to log out in the User Guide in the page about this role.

buttonwillowsix: @pookmish I think this is ready for a re-review. 

Since I don't need to test the toolbar now, the test is just stubbed out... testing nothing. I couldn't think of a good page to test for testing "unpublished." Does this test happen against the initial state? (In which case we could test one of the unpublished bits of default content?)

pookmish (approved):

PR #900, opened 2025-04-04
D8CORE-7924 Add header utility links to site settings

pookmish:
# READY FOR REVIEW

# Summary
- added header links to site settings.

# Review Tasks
1. open gitpod.
2. open site settings page: `/admin/config/system/basic-site-settings`
3. Open the "Site Header Options"
4. Fill button and link fields.
5. save config page
6. go to home page
7. observe the button and the links.


pookmish:
@buttonwillowsix, I'd like to get this merged in before I leave so that I can push it up to the test environment. Then while I'm out next week, any subthemes could be reviewed if there's time.

buttonwillowsix:
@pookmish sure thing! I thought I commented before, but I must have closed that tab before saving. Sorry about that!
I thought it looked great and I like the name "Site Header Optons." (I am going to update the User Guide to use the term "Site Header" more consistently!) 

My only three comments are very minor and could probably be added later in the interest of getting this finalized, style and QA-wise

- I wonder if we need to put it all in a little group. Right now it feels related to the Drop Down menu checkbox, which it isn't.
- I wonder about adding the button last since that is likely to display last.. but we can look at that again after Kerri tweaks the design.
- I think we will need to add some help text (and maybe some guidance about length!)

buttonwillowsix (approved):
Does the job! Yay!

buttonwillowsix:
@pookmish this is set to automerge... that seemed to be the only merge choice!

PR #1022, opened 2025-11-11
D8CORE-8412 | replace spotlight text in layout with content block; show body field on Spotlight variant

rebeccahongsf:
# READY FOR REVIEW

# Summary
- D8CORE-8412 | replace spotlight text in layout with content block; show body field on Spotlight variant

# Review By (Date)
- When possible

# Review Tasks

1. Check out this branch
2. Rebuild Cache and import config `drush cr ; drush ci`
3. Create a News spotlight variant. 
4. Verify that the body field appears in the Spotlight variant form
5. Fill out fields and save.
6. Navigate to that spotlight page in an incognito browser
7. Verify that the "Spotlight" text appears.
8. Review code

- D8CORE-8412

### DIFF
--- a/config/sync/core.entity_form_display.node.stanford_news.default.yml
+++ b/config/sync/core.entity_form_display.node.stanford_news.default.yml
@@ -113,25 +113,7 @@ content:
         maxlength_js_summary: null
         maxlength_js_label_summary: 'Content limited to @limit characters, remaining: <strong>@remaining</strong>'
         maxlength_js_enforce: false
-      conditional_fields:
-        a1b2c3d4-e5f6-7890-a1b2-c3d4e5f67890:
-          entity_type: node
-          bundle: stanford_news
-          dependee: layout_selection
-          settings:
-            state: '!visible'
-            reset: false
-            condition: value
-            grouping: AND
-            values_set: 1
-            value: ''
-            values: {  }
-            value_form:
-              -
-                target_id: news_spotlight
-            effect: show
-            effect_options: {  }
-            selector: ''
+      conditional_fields: {  }
   group_banner:
     weight: 26
     region: content
--- a/config/sync/layout_library.layout.news_spotlight.yml
+++ b/config/sync/layout_library.layout.news_spotlight.yml
@@ -70,7 +70,7 @@ layout:
               field_label:
                 label_value: ''
                 label_tag: ''
-        weight: 10
+        weight: 11
         additional: {  }
       a882d0a7-3cbf-49b9-a5e9-1907b4756679:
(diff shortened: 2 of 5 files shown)


config/sync/core.entity_form_display.node.stanford_news.default.yml
```diff
@@ -113,25 +113,7 @@ content:
...
-            values_set: 1
-            value: ''
-            values: {  }
-            value_form:
-              -
-                target_id: news_spotlight
-            effect: show
-            effect_options: {  }
-            selector: ''
+      conditional_fields: {  }
```
rebeccahongsf, line 116: @pookmish we chatted about keeping the `body` field as a part of spotlights last week. I dropped the ball on updating [Ian's ticket](https://stanfordits.atlassian.net/browse/D8CORE-8390) so I'm adding that `body` field back.

config/sync/layout_library.layout.news_spotlight.yml
```diff
@@ -147,34 +147,21 @@ layout:
...
-        weight: 9
+        weight: 10
         additional: {  }
-      b9778f05-c8e7-4f90-8456-3d0125d36538:
-        uuid: b9778f05-c8e7-4f90-8456-3d0125d36538
+      1067a8ed-6ab9-401a-a6df-fe6c069d4a08:
+        uuid: 1067a8ed-6ab9-401a-a6df-fe6c069d4a08
         region: left
         configuration:
-          id: 'field_block:node:stanford_news:layout_selection'
```
rebeccahongsf, line 156: The layout selection wasn’t appearing for users who are not logged in. I tested a few things, and I’m guessing it’s because the layout selection is limited to logged-in users.

@pookmish would this be a reasonable workaround to adding the "Spotlight" text?

pookmish: Using an inline block like this requires a whole other bunch of work because blocks are "content" while this is configuration. We don't include "content" in code.

What you can do though, is create a "simple block" and place it on the layout. `/admin/structure/block/simple-block`

pookmish: Alternatively, if it can be a heading element (which i don't think it can), you could use the "Heading block"

Either way, this sounds like a good use of an acceptance test.

pookmish (changes requested):
Swap for a non-content block

config/sync/layout_library.layout.news_spotlight.yml
rebeccahongsf: That makes sense to me! Thank you for the direction, Mike 👍 Simple block and acceptance test added!

pookmish (approved):

